PR #5, opened 2021-11-29
Received typography from Design system stakeholders and using tailwin…

KhalidAdan:
### Description

Updates typography and uses tailwind forms to reset form css

N/A

### Addtional Notes
N/A


tailwind.config.js
```diff
@@ -3,66 +3,14 @@ module.exports = {
   purge: ['./pages/**/*.{js,ts,jsx,tsx}', './components/**/*.{js,ts,jsx,tsx}'],
   darkMode: false,
   theme: {
-    screens: {
-      desktop: '1200px',
-      tablet: '992px',
-      mobile: '768px',
```
KhalidAdan: After talking with the design stakeholders I realized we will most likely need to change this information. The design spec is still in the research phase and will most likely change. 

In the meantime I'm going to use tailwind's built in responsive breakpoints and we can override them by name once the designs are closer to final.

PR #6, opened 2021-11-29
discussion topic for handling the dynamic form fields

KhalidAdan:
This is just to serve as the catalyst of a conversation about how the backend will supply the information required to the frontend to render a dynamic form. One missing piece is that the questions are currently nested underneath different headers and the backend will need to supply that information too. For more information on that see SAEB-786 in Jira and take a look at the mockups.


pages/api/calculateEligibility.ts
```diff
@@ -10,6 +10,19 @@ import {
...
+  fieldLabel: string
+  fieldRequired: boolean
+  fieldPlaceholder: string
+  fieldType: 'text' | 'number' | 'email' | 'radio' | 'checkbox'
+}
+
+type allFieldProposal = {
+  // ..other data ...
+  fields: formField[]
+}
```
KhalidAdan, line 24: First pass at what I was thinking, where `allFields` would instead just be exposed as `fields` and we could parse that information into a component factory to create the form.

From there we stuff user inputs into a context (might need a state management library to stop unnecessary re-renders) and submit updates onChange.

PR #19, opened 2021-12-08
Feature/dynamic form

KhalidAdan:
### Description

This PR adds dynamic form rendering, and a working slice of the prototype! The results tab currently has some crude logic to render the correct view which will eventually removed for the summary page.

Getting large so there will be a follow-up PR with cleanup


utils/api/definitions/fields.ts
```diff
@@ -23,11 +23,11 @@ enum FieldType {
 export const fieldDefinitions: FieldDefinitions = {
   [FieldKey.INCOME]: {
     key: FieldKey.INCOME,
-    label: 'What is your current net income?',
+    label: 'What is your current annual net income in Canadian Dollars?',
     category: FieldCategory.INCOME_DETAILS,
     order: 1,
     type: FieldType.NUMBER,
-    placeholder: '20000',
+    placeholder: '$20,000',
```
KhalidAdan, line 30: Updated placeholder in API here, just to match spec a little tighter -- doesn't seem to be any test implication

utils/api/definitions/fields.ts
```diff
@@ -46,22 +46,22 @@ export const fieldDefinitions: FieldDefinitions = {
...
     default: undefined,
   },
+  [FieldKey.PARTNER_INCOME]: {
+    key: FieldKey.PARTNER_INCOME,
+    label: "What is your partner's annual net income in Canadian dollars?",
+    category: FieldCategory.PARTNER_DETAILS,
+    order: 5,
+    type: FieldType.NUMBER,
+    placeholder: '20000',
+  },
```
KhalidAdan, line 64: Updated the ordering of the questions to match mocks, updated tests to match

next-env.d.ts
```diff
@@ -1,5 +1,4 @@
 /// <reference types="next" />
-/// <reference types="next/types/global" />
 /// <reference types="next/image-types/global" />
 
 // NOTE: This file should not be edited
```
KhalidAdan, line 4: Not exactly sure what changed to make this update happen, but Next made it. My guess is when updating the frontend to no longer use `GetServerSideProps` this file was regenerated as we have no exported types from next

components/Tooltip/tooltips.en.json
```diff
@@ -0,0 +1,43 @@
...
+  "yearsInCanadaSince18": [
+    "Years Lived in Canada",
+    "This includes periods when you normally lived in Canada. If you have not lived in Canada all of your life, any absences from Canada longer than 6 months are not included. "
+  ],
+
+  "everLivedSocialCountry": [
+    "Social Agreement Countries",
+    "You may still qualify if you have you lived in one of the countries Canada has established a social security agreement."
+  ]
+}
```
KhalidAdan, line 43: Placeholder for tooltips data just an array for now. In the future I can make this an array of objects keyed by the question and maybe we place it in the global endpoint we chatted about.

components/Layout/index.tsx
```diff
@@ -9,20 +9,19 @@ export const Layout: React.VFC<{
   children: React.ReactNode
 }> = ({ children }) => {
   const otherLang = useInternationalization('otherLang')
-  const otherLangNameFull = useInternationalization('otherLangFull')
```
KhalidAdan, line 12: removing i18n for now,  though it will make it's way back!

components/Forms/ComponentFactory.tsx
```diff
@@ -0,0 +1,261 @@
...
+      if (value == '') {
+        continue
+      }
+      if (qs !== '') qs += '&'
+      qs += `${key}=${value}`
+      formCompletion[key] = value
+    }
+
+    //redirect to exit case if income is too high
+    if (parseInt(formData.get('income') as string) > 129757)
```
JeremyKennedy: I do not like the max income being defined here, but we can remove that later, I hope!
edit: saw it used in two other places, definitely would like a `const` here

components/Forms/ComponentFactory.tsx
```diff
@@ -0,0 +1,261 @@
...
+
+      oas(newFormData.oas)
+      gis(newFormData.gis)
+      allowance(newFormData.allowance)
+      afs(newFormData.afs)
+
+      //set Progress
+      checkCompletion(newFormData.fieldData, formCompletion, setProgress)
+    } else {
+      // handle error
```
JeremyKennedy: to do? 😄

components/Forms/ComponentFactory.tsx
```diff
@@ -0,0 +1,261 @@
...
+                      ...styles,
+                      boxShadow: 'none',
+                    }),
+                  }}
+                  className="rselect"
+                  isSearchable
+                  isClearable
+                  placeholder="Select from..."
+                  defaultValue={
+                    field.key == 'maritalStatus'
```
JeremyKennedy: why is maritalStatus handles specially here?

components/Forms/ComponentFactory.tsx
```diff
@@ -0,0 +1,261 @@
...
+                    if (!newValue) return
+
+                    const form: HTMLFormElement = document.querySelector(
+                      'form[name="ee-form"]'
+                    )
+                    if (!form) return
+
+                    const formData = new FormData(form)
+
+                    // prepare GET request to send to backend
```
JeremyKennedy: This is duplicated code from line 47 and should be moved into a function.

components/Forms/ComponentFactory.tsx
```diff
@@ -0,0 +1,261 @@
...
+                      if (qs !== '') qs += '&'
+                      if (key == field.key) {
+                        qs += `${key}=${newValue.value}`
+                      } else {
+                        qs += `${key}=${value}`
+                      }
+                      formCompletion[key] = value
+                    }
+
+                    fetch(`api/calculateEligibility?${qs}`)
```
JeremyKennedy: again duplicated from line 61

components/Forms/Radio.tsx
```diff
@@ -0,0 +1,63 @@
...
+const correctForBooleans = (value: string) => {
+  switch (value) {
+    case 'Yes':
+      return 'true'
+      break
+    case 'No':
+      return 'false'
+      break
+    default:
+      return value
```
JeremyKennedy, line 60: if we reach this line, does that mean the boolean is invalid? and this should raise an error?

package.json
```diff
@@ -22,15 +22,20 @@
...
   "dependencies": {
+    "@headlessui/react": "^1.4.2",
     "@babel/core": "^7.16.5",
     "@tailwindcss/forms": "^0.3.4",
+    "@types/lodash": "^4.14.177",
     "isarray": "^2.0.5",
     "joi": "^17.5.0",
-    "next": "^12.0.7",
+    "lodash": "^4.17.21",
+    "next": "^12.0.3",
```
JeremyKennedy, line 32: unintentional downgrade of `next`, however I did an intentional downgrade in my yarn 2 branch, so all good here!

pages/eligibility/index.tsx
```diff
@@ -0,0 +1,375 @@
...
+  const { query } = useRouter()
+  const [oas, setOAS] = useState<BenefitResult>(null)
+  const [gis, setGIS] = useState<BenefitResult>(null)
+  const [allowance, setAllowance] = useState<BenefitResult>(null)
+  const [afs, setAFS] = useState<BenefitResult>(null)
+  const [progress, setProgress] = useState({ personal: false, legal: false })
+  const [selectedTabIndex, setSelectedTabIndex] = useState<number>(0)
+
+  // check if income is too high to participate in calculation
+  const incomeTooHigh = query && parseInt(query.income as string) > 129757
```
JeremyKennedy: really not liking defining eligibility logic here, especially that it is defined in 3 different places, at a minimum it should be broken out into a `const` so there is only one place to update it

JeremyKennedy (changes requested):
Few small changes I'd really like done before merging (like the income `const` and perhaps the duplicated code) but the other stuff is fine.

components/Forms/ComponentFactory.tsx
KhalidAdan: Haha, yeah -- Alla needs to get me the designs for the validation labels so there is a blocker here

components/Forms/ComponentFactory.tsx
KhalidAdan: My bad, I just put this put with the intention of cleanup coming in with my next PR, but I don't mind doing that now. Thought you wanted to merge so we didn't have lot's of conflicts!

components/Forms/ComponentFactory.tsx
KhalidAdan: I learned in a meeting before showing the app to Dave and Amanda that the default is not always the 0th element in the array, this was a hacky way to get it to work for the demo! Totally not final!

components/Forms/Radio.tsx
KhalidAdan: No because this function is not called only for booleans. This question might also be rendering a regular radio question, in which you should just return the default value.

I guess technically to get rid of this check altogether we should never have boolean typed questions and they should be a radio with the values of 'Yes' and 'No'. 

Or maybe this is a checkbox question? Would require us to run it by Alla and Joan, what do you think?

JeremyKennedy (approved):
love it! my last feedback is that I would prefer to squash this in, rather than merge, because the commit history is quite a mess with all the merge commits. but if you want to preserve the history for your use, then that's cool too!

PR #21, opened 2021-12-13
Merge develop to main

JeremyKennedy:
- updated legal status to use new options
- citizen, permanent non sponsored, and indian are all treated the same
- sponsored will return conditional eligibility and to contact service canada (not sure if this is correct?)
- other will return ineligible due to no legal status in canada (I now realize this should say to contact service canada, will change that tomorrow)
- when client is under 65 and part of an agreement country, we will now say (paraphrased) "you are ineligible, but you may be eligible when you turn 65, depending on the agreement" instead of saying they are conditionally eligible
- for OAS when age is 64, they are now directed to service canada (but still ineligible)
- reword Survivor ineligibility due to marital status to "You must be a surviving partner or widowed to be eligible for Allowance for Survivor'
- fixed typo: Seperated -> Separated (you will need to update tests/documentation)
- for all conditional results, the text has been updated to encourage contacting service canada
- the backend now provides the frontend the full set of fields to display and the relevant configurations
- the backend now clearly defines the order of fields (and all responses are sorted accordingly)
- sponsored status now returns conditional for all benefits
- Allowance and Survivor have been updated to match most of the improvements made to OAS (see first three points from previous message)
- now, after providing income, most other fields will be required right away, reducing pop-in
- tweaked/improved ordering and categorization for fields
- change marital status to dropdown


__tests__/pages/api/index.test.ts
```diff
@@ -276,310 +291,345 @@ describe('basic OAS scenarios', () => {
...
+      income: 10000,
+      age: 64,
+      livingCountry: LivingCountry.CANADA,
+      legalStatus: LegalStatus.CANADIAN_CITIZEN,
+      yearsInCanadaSince18: 9,
+      everLivedSocialCountry: true,
+    })
+    expect(res.body.oas.eligibilityResult).toEqual(ResultKey.INELIGIBLE)
+    expect(res.body.oas.reason).toEqual(ResultReason.AGE)
   })
```
KhalidAdan, line 432: these are some super thorough tests, good job!

utils/api/benefits/checkAfs.ts
```diff
@@ -0,0 +1,156 @@
...
+      eligibilityResult: ResultKey.INELIGIBLE,
+      entitlementResult: 0,
+      reason: ResultReason.SOCIAL_AGREEMENT,
+      detail:
+        'You currently do not appear to be eligible for the Allowance for Survivor as you have indicated that you have not lived in Canada for the minimum period of time or lived in a country that Canada has a social security agreement with. However, you may be in the future if you reside in Canada for the minimum required number of years.',
+    }
+  }
+  // fallback
+  throw new Error('should not be here')
+}
```
KhalidAdan, line 156: Looks to be some moved around logic, correct? Not all net new?

utils/api/definitions/fields.ts
```diff
@@ -0,0 +1,133 @@
...
+
+interface FieldDataDropdown extends FieldDataGeneric {
+  type: FieldType.DROPDOWN
+  values: Array<string>
+  default?: string
+}
+
+type FieldDefinitions = {
+  [x in FieldKey]: FieldData
+}
```
KhalidAdan, line 133: 👍🏾

KhalidAdan (approved):
Looks good! I'll be changing most of the front end before the demo so I've gone ahead and ignored it for now.

__tests__/pages/api/index.test.ts
JeremyKennedy: yeah, tbh this is my first project with tests, haha! they are a bunch of them but I'm definitely missing some coverage... but is good enough that in general when I break something, the tests catch it, which is the whole point (and I'm really liking working that way)!

utils/api/benefits/checkAfs.ts
JeremyKennedy: yeah I moved a lot of stuff around, I think most of the changes in these files is refactoring/cleaning up and not new logic. I tried to organize all the conditions a bit more logically, with the "eligible" path at the top, and the rest is just determining reason for ineligibility.

utils/api/definitions/fields.ts
JeremyKennedy: haha yeah fun stuff, it makes sure that I don't forget to add the field definition whenever I add a new field! I love typescript!

PR #28, opened 2021-12-14
Update Yarn, use its new features

JeremyKennedy:
There is a lot going on here. I would review this by commit, rather than looking at all files at once.

The first commit is a pretty straightforward transition to Yarn 2. [Why?](https://yarnpkg.com/getting-started/qa#why-should-you-upgrade-to-yarn-modern)

The second commit simply updates some commands to use Yarn explicitly, not too exciting.

The third uses the Plug and Play feature of Yarn 2, this gets rid of `node_modules` completely and enforces properly specified dependencies. It should speed things up? See [here](https://yarnpkg.com/features/pnp) for more info. Unfortunately many of our dependencies don't specify their own dependencies in a proper way, so our own `package.json` is now littered with the dependencies of other projects. It's ugly, but it works.

The last commit is pretty "fun". It uses the "zero install" approach of Yarn 2, where all dependencies are committed to the repo, and the idea is you can switch between branches with different dependencies, without having to `yarn install`. More info [here](https://yarnpkg.com/features/zero-installs).

As far as I can tell, nothing is broken. This will definitely conflict with your branch in progress. I think we should merge your stuff in, and then I will rebase this on top of it.

Any thoughts on this? Love it (new stuff, yay!)? Hate it (change!!!)? Maybe Yarn2=good but PnP=bad? Let me know what you think!


JeremyKennedy:
Looks like Cypress is angry, it is not something I tested. Before merging this, I should fix that.

JeremyKennedy:
Closing this, will be opening PRs for each step (basic, PnP, zero-install).

PR #36, opened 2021-12-15
Yarn 2: Plug and Play

JeremyKennedy:
This uses the Plug and Play feature of Yarn 2, this gets rid of `node_modules` completely and enforces properly specified dependencies. It should speed things up? See [here](https://yarnpkg.com/features/pnp) for more info. Unfortunately many of our dependencies don't specify their own dependencies in a proper way, so our own `package.json` is now littered with the dependencies of other projects. It's ugly, but it works.

As far as I can tell, nothing is broken.

Any thoughts on this? Love it (new stuff, yay!)? Hate it (change!!!)? Maybe Yarn2=good but PnP=bad? Let me know what you think!


JeremyKennedy:
Closing mainly due to complexity with not much added value.

PR #37, opened 2021-12-15
Yarn 2: Zero-install

JeremyKennedy:
This one is pretty "fun". It uses the "zero install" approach of Yarn 2, where all dependencies are committed to the repo, and the idea is you can switch between branches with different dependencies, without having to `yarn install`. More info [here](https://yarnpkg.com/features/zero-installs).

As far as I can tell, nothing is broken.

Any thoughts on this? Love it (new stuff, yay!)? Hate it (change!!!)? Maybe Yarn2=good but zero-install=bad? Let me know what you think!


JeremyKennedy:
Closing mainly due to complexity with not much added value.

PR #38, opened 2021-12-15
Add GIS entitlement

JeremyKennedy:
Still need to ask Partner Age, so that we can determine whether to use table 3 or 4.

### DIFF
--- a/utils/api/benefits/checkGis.ts
+++ b/utils/api/benefits/checkGis.ts
@@ -54,13 +54,14 @@ export default function checkGis(
             'You may be eligible, please contact Service Canada for more information.',
         }
       } else {
+        const entitlementResult = new GisEntitlement(
+          value.income,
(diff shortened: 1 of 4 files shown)


utils/api/benefits/checkGis.ts
```diff
@@ -7,6 +7,8 @@ import {
 import { GisSchema } from '../definitions/schemas'
 import { BenefitResult, CalculationInput } from '../definitions/types'
 import { validateRequestForBenefit } from '../helpers/validator'
+import { OutputItem } from '../scrapers/_base'
+import gisTables from '../scrapers/output'
```
KhalidAdan, line 11: interesting, the source of truth for GIS resides in tables on the web somewhere?

KhalidAdan (approved):
Looks like a clean solution! Just a quick question on my end.

Looks good!

PR #47, opened 2021-12-20
Feature/minor improvements

JeremyKennedy:
- add openapi tests
- when does not meet legal status or age, return ineligible instead of conditional
- rename "ineligible" to "not eligible"
- update partner income placeholder
- exclude "Agreement" from list of countries
- add examples for other legal status
- .idea: rename a file
- separate oas entitlement into own test suite


KhalidAdan (approved):
Nice! Breaking out OAS into it's own suite helps for searchability, and the copy changes look good!

PR #48, opened 2021-12-20
Attempt to fix dependabot

JeremyKennedy:
No idea if this will work... if it doesn't, well, then we've updated some dependencies.


KhalidAdan (approved):
🙏🏾 🤞🏾

PR #49, opened 2021-12-20
Remove cypress

JeremyKennedy:
Pretty straightforward, I think. Just need a quick glance at the dependencies I've removed. I removed a couple in addition to cypress but I think it's all good. Of course, if you notice any more to remove, let me know!


KhalidAdan (approved):
Can just merge this in. Looks straightforward to me!

PR #60, opened 2022-01-05
Disable Dependabot

JeremyKennedy:
Unfortunately, Dependabot does not support Yarn v2: https://github.com/dependabot/dependabot-core/issues/1297

This change should disable it. 

Here is a working alternative, but I have a feeling our GitHub Org will be hesitant to grant access to this bot: https://github.com/renovatebot/renovate

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,3 +8,4 @@ updates:
     reviewers:
       - JeremyKennedy
       - KhalidAdan
+    open-pull-requests-limit: 0


KhalidAdan (approved):

KhalidAdan:
>Unfortunately, Dependabot does not support Yarn v2: dependabot/dependabot-core#1297
>
>This change should disable it.

I'm all for disabling dependabot

s-laugh (changes requested):
I specifically enabled the dependabot in the settings; it would be good to understand the issues around this and manage this properly.

JeremyKennedy:
The issues are that it does not support the latest stable version of Yarn (see link in PR description). The changes it makes in the PR break the lockfile and therefore the entire deployment/app. You can see this by reviewing any of the open Dependabot PRs.

s-laugh:
Okay, my main concern with flagging this was it seemed like we are removing a security feature that's used to notify us of problems with dependencies. But as I take a closer look - you just don't want it to auto create PRs because they break. However the notification piece can stay. Do I have that right?

JeremyKennedy:
> it seemed like we are removing a security feature that's used to notify us of problems with dependencies. 

I'm all for this feature, but unfortunately Dependabot still does not support Yarn 2 which was released early 2020.  Same with Yarn 3 which was mid 2021. The issue was raised mid 2019, so it's been a problem for a while!
> But as I take a closer look - you just don't want it to auto create PRs because they break. 

Correct, we don't want Dependabot because it is broken on the latest stable Yarn releases.
> However the notification piece can stay. 

Not sure where the notifications would go if not a PR, maybe there's a mechanism I'm not aware of? But yeah, sure! It's just that the PR spam and effort is annoying.
> Do I have that right?

The nicest solution would be to grant the Renovate bot access to our repo (it is essentially a Dependabot that supports Yarn), but as I mentioned initially, I don't expect that to happen. If it can, great! It is the best solution imo if you look through the issue discussion.
Another option would be to keep Dependabot working as-is (broken) and manually fixing every PR it creates. I did that yesterday.
And the last option would be to just stop Dependabot from creating PRs. Or maybe limit it to one, so we still get something, it just won't become overwhelming (but will still need a manual fix for every PR).

s-laugh:
> Not sure where the notifications would go if not a PR, maybe there's a mechanism I'm not aware of? But yeah, sure! It's just that the PR spam and effort is annoying.

Great, here's what I'd like to suggest then. I can disable the Dependabot PR rights in the settings and you wouldn't need this file at all anymore.
image
I'll leave the alerts on; they should show up here https://github.com/DTS-STN/eligibility-estimator/security/dependabot and I believe sometimes on the home page of the repo.

Sound good?

JeremyKennedy:
Yeah that works! Thanks!

s-laugh:
Done

PR #68, opened 2022-01-05
Summary Object: Provide summary of current state to frontend

JeremyKennedy:
Pretty basic right now. It will scan the four benefit results determine the state based on priority:
1. If anything is "needs info", the summary state will be needs info
2. Next is "conditionally eligible" which will lead to "unavailable"
3. Next is "eligible" which will lead to "available: eligible"
4. Last is when nothing matches any of the above, it will lead to "available: ineligible"

I'm thinking you can add support on your end, and then we can play around with it to see what I need to improve on my end.

- [x] Frontend support @KhalidAdan 
- [x] Play with it and review next steps
- [x] Tests @JeremyKennedy


KhalidAdan:
Seems to work well on my end and I got to delete a lotttt of branching code which feels amazing. Give it a whirl and let me know!

KhalidAdan (approved):
Good to go once copy is in!

PR #71, opened 2022-01-06
Add French translations

JeremyKennedy:
- [x] frontend: support passing `_language` in query string (optional parameter, defaults to `EN`)
- [x] frontend: support `{ key: string, text: string }` format for question categories
- [x] frontend: support `{ key: string, text: string }` format for question options aka field values
- [ ] copy proper/correct translations (waiting on those)
- [x] review excel spreadsheet of countries (Jeremy with Wiam probably)

I did the key/text stuff because these pieces of data are used not only for displaying text to the user, but also for logic. For example, on your end you have logic depending on the question category, you will want to depend on the new `key` rather than the original `text`. Another example is on my end, receiving the requests, I will be expecting the `key` and not the `text`.


JeremyKennedy:
Did a full review, looks good! We still need to apply the new translations. Can do that after merging this.

KhalidAdan (approved):
This looks good! I've tinkered a bunch and it seems like it works fine to me!

PR #73, opened 2022-01-09
Feature/client state mgmt

KhalidAdan:
This branch introduces a bunch of fixes, listed as:

- Clear button now clears the results from the results tab
- Call to action text now exists when results are available
- Call to action has a link to the FAQ tab
- Home page's clear button also works
- Introduces `mobx-state-tree` as the client state management tool
    - Removes almost all of `React`'s state management and replaces them with observers and `mobx-state-tree` models
- Cuts `React`'s re-renders down immensely with no loss in app responsiveness 🎉 🎉 🎉 
- Other miscellaneous cleanup and performance fixes


components/Forms/Input.tsx
```diff
@@ -14,53 +17,80 @@ export interface InputProps extends InputHTMLAttributes<HTMLInputElement> {
...
+          onChange={onChange}
+          required
         />
-      ) : (
-        <input
-          name={props.name}
-          data-testid={props.name}
-          {...props}
+      )}
+      {(name == FieldKey.AGE || name == FieldKey.YEARS_IN_CANADA_SINCE_18) && (
```
KhalidAdan: Not urgent but this is a little bit of a bummer, but needs to exist because a number type is a little too vague unfortunately. 

| Field                   | Issue / example                                                            |  
|------------------------|----------------------------------------------------------------------------|
| Age                    | just a regular number                                                      |
| Income, Partner Income | Requires specific filetering to add a `thousandSeparator` and a $ `prefix` | 
| Other                  | text, not a number at all                                                 

Could we add a currency type for `income` and `partnerIncome`? That would mean I could just use the type and render one three new input fields.

e.g. `TextField`, `NumberField`, `CurrencyField` 

Works as is now, but would definitely remove a lot of code on my end.

components/Forms/Radio.tsx
```diff
@@ -1,61 +1,75 @@
...
  * A form input radio field that is rendered by the component factory.
  * @param props {InputProps}
  * @returns
  */
-export const Radio: React.VFC<InputProps> = (props) => {
-  const { query } = useRouter()
-
+export const Radio: React.VFC<InputProps> = observer((props) => {
+  const { name, label, checkedValue, onChange, values, keyforid, error } = props
+  console.log(checkedValue)
```
KhalidAdan: remove

components/Forms/Input.tsx
JeremyKennedy: Applied my changes to satisfy this!

components/Forms/Radio.tsx
JeremyKennedy: you commented remove, but it is still here?

client-state/models/Form.ts
```diff
@@ -0,0 +1,206 @@
...
+      parent.setSummary({})
+    },
+    clearAllErrors() {
+      self.fields.map((field) => field.setError(undefined))
+    },
+    setupForm(data: FieldData[]): void {
+      data.map((fieldData) => {
+        const field = self.getFieldByKey(fieldData?.key)
+
+        const defaultValue = (fieldData as any).default
```
JeremyKennedy: I'm curious why you need to use `as any` here, shouldn't `fieldData` have the `FieldData` type with the properties you're accessing?

client-state/models/FormField.ts
```diff
@@ -0,0 +1,74 @@
...
+      self.error = error
+    },
+    clearValue() {
+      self.value = null
+    },
+    sanitizeInput() {
+      let val = ''
+
+      if (KeyValue.is(self.value)) {
+        if (self.key == FieldKey.LIVING_COUNTRY || FieldKey.MARITAL_STATUS) {
```
JeremyKennedy: why are you treating LivingCountry and MaritalStatus specially here? And why is LegalStatus not treated the same? Can't you use the type property instead of having logic specific to the fieldKey?

JeremyKennedy: Oh I guess LegalStatus is different because it's radio and not dropdown

JeremyKennedy (approved):
Looks good, I added a couple curiosity questions which do not need to be addressed (as long as you can justify what you have!), and we can remove the `console.log` lines later since I'm sure there's more unnecessary ones!

client-state/models/FormField.ts
KhalidAdan: Yeah legal status being a radio means I can just use the value provided but the other two need to be massaged into the format required by the API before sending them, but this might be a moot discussion because I don't think I need those anyway! 

The `KeyValue.is()` check should give me all I need and the API will always want the key, so I can remove that if entirely! Good eye!

client-state/models/Form.ts
KhalidAdan: Hmm good point, I was just getting around the default being undefined some times but I can do this a different way. I'll make the change!

KhalidAdan: fixed with
```js
let defaultValue = null
if ('default' in fieldData) {
  defaultValue = fieldData.default
}
```

client-state/models/FormField.ts
KhalidAdan: fixed with

```js
if (KeyValue.is(self.value)) {
  val = self.value.key
} else {
  if (self.value.includes('$')) {
    val = self.value.toString().replace('$', '').replace(',', '')
  } else {
    val = self.value.toString()
  }
}
```

PR #77, opened 2022-01-10
Adjust Reason Priorities for Allowances

JeremyKennedy:
Decreases the priority of the "income too high" reason for ineligibility. Now, will say age is the reason when both age and income do not satisfy. 

Do not merge yet as I'm waiting on a clarification from Joan, but I'll still take the approval!

- [x] Clarify: when age and partnerReceivingOas do not satisfy, what should be the reason?


KhalidAdan (approved):
Makes sense to me 🚢

PR #78, opened 2022-01-10
Backend Wording Improvements

JeremyKennedy:
Hopefully this satisfies everyone who complains about our wording!!

- Copied all text from the spreadsheet, including French translations.
- Added question category "Social Agreement Countries"
- Moved partner questions to bottom
- Add link to social agreement countries within the question text
- When Partial OAS, the reason detail will now mention that it is Partial OAS. The reason key is updated as well (and therefore all tests)

- [ ] Support HTML in question text (need to merge other branch) @KhalidAdan 
- [ ] Update "contact service canada" links to actually be links @JeremyKennedy


KhalidAdan (approved):
Ordering and copy looks good! The polish is coming along nicely!

KhalidAdan:
~Starting on my end!~

Edit: I think it would be better if we just merge this into develop once you are done, as the HTML questions should just work. Even if they don't, client-state-management will ensure they do!

PR #82, opened 2022-01-12
Fixed an issue where an empty array was returning a filled out progress due to the idiosyncrasies of Array.every()

KhalidAdan:
As the title says! 

```js
[].every(field => field.filled) // this returns true!  
```


JeremyKennedy (approved):
what a silly behavior! approved!

PR #83, opened 2022-01-12
Bugfix/ux design pass

KhalidAdan:
Alla and I did a design pass to make sure the app was indicative of the vision, and we decided on a number of changes. Most of them were harmonizing margins but the full list is below:

- Marital status should not have a clearable or searchable function
- Radio button spacing felt a little claustrophobic, fixed
- Label were all harmonized to consistent margins
- Hidden questions are now rendered with full width
- Call to action's margins have been updated
- More information had a bug where tailwind applied it's own margin, forced it to be correct
- Unavailable results will now render a full width picture
- Error labels now have a consistent amount of spacing in the app
- The alerts will now responsively render their SVGs

### DIFF
--- a/components/Alert/index.tsx
+++ b/components/Alert/index.tsx
@@ -1,13 +1,25 @@
 import { useRouter } from 'next/router'
-import React from 'react'
+import React, { useEffect, useRef, useState } from 'react'
 import { EstimationSummaryState } from '../../utils/api/definitions/enums'
+
+/** offset padding and margin of the component, used to calculate svg height responsively */
+const OFFSET_WIDTH = 28
 
 export const Alert: React.VFC<{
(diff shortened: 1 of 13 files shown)


JeremyKennedy (changes requested):
My thoughts:

- [x] The radio button change went too far, imo the circles are too big and a nice compromise would be somewhere in the middle of before and now (edit: discussed that this was already a tough sell and compromise so I'll accept it as-is)
- [x] The marital status dropdown is now shorter (height) than the other fields - it should be the same
- [ ] Ready to merge once the new DROPDOWN_SEARCHABLE field type is used, rather than checking for the field key == marital status.

JeremyKennedy (approved):

PR #89, opened 2022-01-13
Add new "Partner Benefit Status" question, plus major backend refactor

JeremyKennedy:
I changed a lot here, let's break it down:

Functional/logic/main changes:

- Add new partnerBenefitStatus question. This affects GIS eligibility/entitlement, and allowance entitlement.
- Specifically, now Allowance will require partnerBenefitStatus of GIS.
- Also, now GIS entitlement will give up (return -1) when either the client or the partner is getting only Partial OAS. The reasoning text will say that entitlement could not be determined.
- Last one relating to partnerBenefitStatus, now when partnerBenefitStatus is Allowance, the correct table will be used for GIS entitlement.
- When the OAS entitlement result is partial, the reason detail text will mention this.
- The text "conditionally eligible" is changed to "unavailable".

To do:

- [x] Need to properly handle unavailable entitlements, specifically the -1 GIS entitlement when partnerBenefitStatus is partial @KhalidAdan 

Non-functional/refactoring/transparent changes:
- Centralized a lot of logic into a RequestHandler class, including a bunch of duplicate stuff. Simplified a lot in doing this. This gathered field requirements, field data building, result calculations, normalization, and probably more. Essentially everything that existed in each benefit is now moved up into a single helper class.
- The returned object from the API has moved each benefit's results into a results property, rather than existing right in the root of the object (check the ResponseSuccess type)
- With the change to field requirements, now when the result is moreInfo, the returned ResponseSuccess will have an empty results object. This is because now the summary object will handle the moreInfo state completely.
- Added helpers for several fields (marital, legal, country, partner), centralizing logic for those too. So rather than having a marital status check for single/divorced/widowed/separated in each benefit, now it's in one place and checked by `maritalStatus.single`. Cool!
- Reordered fields consistently in all code, tests, openapi/insomnia, etc.
- Synced openapi/insomnia with recent changes
- I've, from what I can tell, updated the necessary places in the frontend to accommodate everything in this PR.
- Started a big cleanup of the tests. The plan is to remove all the redundant tests, as there are a ton of tests that have the exact same input, but only check certain things. Meaning, we could have the same input for four tests, with each one validating the results for one benefit. This makes maintenance pretty sucky so I'm hoping to consolidate all the duplicate ones. Not gonna include that in this PR though.


KhalidAdan (approved):
I think we can go ahead with this!

PR #97, opened 2022-01-17
Add conditional links

JeremyKennedy:
This PR seeks to add conditional links to the UI (SAEB-854). For example, if the client is eligible for Partial OAS, we should display a link about Partial OAS.

The "Links" tab in our main spreadsheet contains the rules for this.

Todo:
- [x] Frontend support @KhalidAdan. Note that this should be the main source for links for the frontend now (assuming we are putting static and conditional links in the same place)

The conditional logic I came up with on the spreadsheet definitely needs some improvements.

### DIFF
--- a/i18n/api/en.ts
+++ b/i18n/api/en.ts
@@ -334,5 +334,87 @@ const en: Translations = {
     availableIneligible:
       'Based on the information you have provided, you are likely not eligible for any benefits. See the details below for more information.',
   },
+  links: {
+    oasOverview: {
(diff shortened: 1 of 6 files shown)


KhalidAdan:
So I was already iterating over the links array before when it was just a test value so I think this work is done already! Feel free to confirm but I just did and it looks good!

image

KhalidAdan (approved):
Looks good! I'm sure we'll get more links in the future and this seems like a pretty painless way to add / update them!

PR #99, opened 2022-01-18
Bugfix/ux and misc polish changes

KhalidAdan:
This PR includes:
- responsive changes for the progress bar and the results table components
- Addition of new breadcrumb design **MOBILE**
- Removal of a debounce in the text/currency/number fields that made the internal state only after 1s
- Addition of new form button layout according to mobile designs **MOBILE**
- Overhaul of the padding and margins to be more consistent
- Radio buttons appear larger and have more padding 
- Other text field is now a textarea with exactly three rows
- Number fields are now constrained to 3 characters of width
- New app name has been added
- Tooltips appear wider in desktop mode and the paragraph padding has been updated to read better **MOBILE and DESKTOP**


JeremyKennedy (changes requested):
- responsive changes for the progress bar and the results table components
    - I see the total row has been added to the results table, and the progress bar margins/spacing is updated
- Addition of new breadcrumb design
    - good!
- Removal of a debounce in the text/currency/number fields that made the internal state only after 1s
    - seems great! I was able to press enter on the last field and be brought to the results page, rather than the field being cleared and getting a requirement error.
- Addition of new form button layout according to mobile designs
    - confirmed, but don't love it!
- Overhaul of the padding and margins to be more consistent
    - on mobile the summary box margins and layout is messed up
    - tooltip button is cut off on mobile, oddly switching to then from desktop fixes this
    - oddly the cause of this one is switching to then from desktop, so it could be argued as low prio, but there is some global margin change having a negative impact on most of the app, specifically the right side of mobile's blue header at the top, results tab table headers, and question tab padding
- Radio buttons appear larger and have more padding
    - holy cow I did not realize how bad this used to be on mobile lol
- Other text field is now a textarea with exactly three rows
    - confirmed, but don't love it!
- Number fields are now constrained to 3 characters of width
    - confirmed, but don't love it!
- New app name has been added
    - great!
- Tooltips appear wider in desktop mode and the paragraph padding has been updated to read better
    - confirmed!

To-do:
- [x] on mobile the summary box margins and layout is messed up
- [x] tooltip button is cut off on mobile, oddly switching to then from desktop fixes this
- [x] oddly the cause of this one is switching to then from desktop, so it could be argued as low prio, but there is some global margin change having a negative impact on most of the app, specifically the right side of mobile's blue header at the top, results tab table headers, and question tab padding

JeremyKennedy (approved):
My primary concerns have all been resolved by the recent commits.

Great work!

PR #101, opened 2022-01-18
Added SSR and SSG

KhalidAdan:
Adds serverside rendering and static site generation to the app! Pretty straightforward except for the eligibility smoke test being broken by a lack of props (Next passes serverside rendered data in as a prop). Ended up solving it by doing what `getServersideProps` does and passed that response body in.

### DIFF
--- a/__tests__/pages/eligibility.test.tsx
+++ b/__tests__/pages/eligibility.test.tsx
@@ -7,6 +7,8 @@ import * as nextRouter from 'next/router'
 import React from 'react'
 import Eligibility from '../../pages/eligibility/index'
 import { LanguageProvider, StoreProvider } from '../../components/Contexts'
+import { ResponseSuccess } from '../../utils/api/definitions/types'
+import { mockPartialGetRequest } from './api/factory'
 
 describe('index page', () => {
   let useRouter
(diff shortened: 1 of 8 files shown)


__tests__/pages/eligibility.test.tsx
```diff
@@ -16,16 +18,20 @@ describe('index page', () => {
...
       pathname: '/eligibility?income=20000',
-      query: '',
+      query: { income: '20000' },
       asPath: '',
     }))
   })
 
   it('should render the eligibility page', async () => {
+    const res = await mockPartialGetRequest({
+      income: '20000' as unknown as number,
```
JeremyKennedy: could just change this to 20000 and take out the type stuff? Seems to work on my end when I try it.

pages/eligibility/index.tsx
```diff
@@ -230,4 +215,25 @@ const Eligibility: NextPage = () => {
...
+export const getServerSideProps: GetServerSideProps = async (context) => {
+  const query = context.query
+
+  const host = context.req.headers.host
+
+  const params = Object.keys(query)
+    .map((key) => key + '=' + query[key])
+    .join('&')
+
+  const path = `http://${host}/api/calculateEligibility?${params && params}`
```
JeremyKennedy: I tried changing this to just `${params}` (without the `&&`) and it seemed to work the same, so what's the reason behind this? Not sure what it's doing here.

JeremyKennedy (changes requested):
Some teeny questions, but overall I am liking the effects of this PR! Cool stuff!

pages/eligibility/index.tsx
KhalidAdan: I was under the (incorrect) impression that if params was undefined it would interact badly with the string interpolation going on. To counteract this, I just did a truthiness check that would make sure it worked then append the string. This was some major overengineering on my part and was how I had it with client side rendering.

Fixed in my next commit!

JeremyKennedy (approved):

PR #102, opened 2022-01-18
Add "Have you lived in Canada your whole life" question

JeremyKennedy:
- add new canadaWholeLife question
- updated tests accordingly
- now yearsInCanadaSince18 is only allowed when canadaWholeLife is false
- cleaned up field requirement tests
- add test ensuring translations exist for all questions
- update frontend tooltip

Quick testing shows this works as expected, including in the frontend. 

There is still the issue when answering the conditional question, and then changing a previous answer so the conditional question is hidden. This is reported in the Excel issue log (the marital status one). We should discuss how to handle that error - I could do it on my end, where I purge any fields I don't need, or you could do it on your end, and not send requests to the API with fields that are not required. The latter is tricky, because you can only figure out which fields are required by sending a request to the API... but you won't know which fields to remove or not. So.... to discuss!

Anyway, we can still review/approve/merge this PR even with that conditional question issue. Just wanted to mention it.


KhalidAdan (approved):
Looks good! I'll be free to chat about where we handle the visible fields issue today!

PR #105, opened 2022-01-19
Improve unexpected field handling

JeremyKennedy:
On my end, I changed my logic so that in all cases I'm aware of, things are handled as expected even when unexpected fields are sent to the API. Specifically, when partner income is sent while marital status is single, things are still good.

Also removed the forbidden field rules, so when formerly forbidden fields are sent, there is no longer any errors.


KhalidAdan (approved):
Tested in the app, works VERY well! woo!

PR #108, opened 2022-01-20
Feature/front end tests

KhalidAdan:
This is the first pass at getting better front-end code coverage, it's getting a little large so I think I'll get this half reviewed and then setup the next set of tests in a separate PR.


.gitignore
```diff
@@ -48,3 +48,6 @@ yarn-error.log*
 !.yarn/releases
 !.yarn/sdks
 !.yarn/versions
+
+.vscode/launch.json
+.vscode/settings.json
```
KhalidAdan, line 53: Have some pretty specific launch configuration, and don't want to potentially collide with anything you have!

KhalidAdan:
There seems to be an issue where `toString().replaceAll(x,y)` doesn't exist in the test environment but works fine locally -- trying to fix this now!

Edit: Our version of node on this project does not have replaceAll(), gonna try and use a shim in test


Edit x2: That did it!

JeremyKennedy (approved):
Seems fine to me!

PR #111, opened 2022-01-22
New partner questions, along with major refactor to support

JeremyKennedy:
- add new partner questions
- proper partner handling
- benefit processing refactor
- rename allowance to alw


client-state/store.ts
```diff
@@ -49,63 +58,27 @@ export const RootStore = types
...
-      input: ModelCreationType<
-        ExtractCFromProps<{
-          eligibilityResult: IMaybe<ISimpleType<ResultKey>>
-          entitlementResult: IMaybe<ISimpleType<number>>
-          reason: IMaybe<ISimpleType<string>>
-          detail: IMaybe<ISimpleType<string>>
-        }>
-      >
-    ) {
+    setOAS(input) {
```
KhalidAdan, line 72: the types here are massive and not really developer friendly, so I removed them. I don't think there will be an issue moving forward, as `mobx` ensures we are type safe here and complains loudly if not.

In case you are curious, here is a _part_ of the type I have removed.

```ts
input: IType<ModelCreationType<ExtractCFromProps<{ eligibility: IMaybe<IModelType<{ detail: IMaybe<ISimpleType<string>>; reason: IMaybe<ISimpleType<string>>; result: IMaybe<ISimpleType<ResultKey>>; }, {}, _NotCustomized, _NotCustomized>>; entitlement: IMaybe<...>; }>>, ModelSnapshotType<...>, ModelInstanceType<...>>.create(snapshot?: ModelCreationType<ExtractCFromProps<{
    eligibility: IMaybe<IModelType<{
        detail: IMaybe<ISimpleType<string>>;
        reason: IMaybe<ISimpleType<string>>;
        result: IMaybe<ISimpleType<ResultKey>>;
    }>, ... more type narrowing ...>;
    entitlement: IMaybe<... more types ...>;
}>>
```

JeremyKennedy: weird, can you not define one type and use it for all benefits, like I've done with `types.ts`? Not a huge problem, but it's nice having types be able to catch errors.

KhalidAdan: I'm okay with either approach, though even without typescript to catch an error, if the `create` call is passed the wrong data, it complains very loudly and 500's the page. MST is using TS underneath, and handles things like that for us.

__tests__/pages/api/index.test.ts
```diff
@@ -2260,10 +2975,145 @@ describe('thorough extras', () => {
...
+    expect(res.body.results.gis.eligibility.reason).toEqual(ResultReason.INCOME)
+    res = await mockGetRequest({
+      ...input,
+      income: 25727,
+    })
+    expect(res.body.results.oas.eligibility.result).toEqual(ResultKey.ELIGIBLE)
+    expect(res.body.results.oas.eligibility.reason).toEqual(ResultReason.NONE)
+    expect(res.body.results.gis.eligibility.result).toEqual(ResultKey.ELIGIBLE)
+    expect(res.body.results.gis.eligibility.reason).toEqual(ResultReason.NONE)
+    expect(res.body.results.gis.entitlement.result).toEqual(0.33)
```
KhalidAdan: This test suite is so thorough, huge thumbs up from me!

KhalidAdan (approved):

__tests__/pages/api/index.test.ts
JeremyKennedy: oh, that reminds me, I need to add more tests!

utils/api/helpers/requestHandler.ts
```diff
@@ -273,6 +271,9 @@ export class RequestHandler {
       // Save the result to the client's partnerBenefitStatus field, which is used TBD
       input.client.partnerBenefitStatus.gisResultEligibility =
         partnerGis.eligibility
+      // Save the result to the partner's partnerBenefitStatus field, which is used for partner's alw eligibility
+      input.partner.partnerBenefitStatus.gisResultEligibility =
+        clientGis.eligibility
     }
```
KhalidAdan, line 280: Was this the bug? In isolation I'm not sure what the issue was.

KhalidAdan (approved):
Just a question from me, about new changes

utils/api/helpers/requestHandler.ts
JeremyKennedy: haha yeah, it was a tricky one. the issue was that the client's GIS entitlement depends on the partner's ALW eligibility, which depends on the client's GIS eligibility, (which depends on the client's OAS eligibility) ...... and I was not saving the client's GIS eligibility into the object that determines the partner's ALW eligibility. It's a huge mouthful and it took a while to finally understand it, hence why I put a ton of comments explaining why I'm doing each update!

PR #114, opened 2022-01-25
Prototype!

KhalidAdan:
This PR incorporates all of the quick wins we spoke about in our meeting yesterday and removes the income field from the front page!! Woo!

I'm specifically looking for feedback on what you think of the sticky Need help div, is it positioned okay? (I moved it to the right so It wouldn't over lap the form but lmk what you think about positioning)

You can use the list below to test if you'd like!

- [x] add dynamic links on questions tab 
- [x] "sticky" need help dialog on questions tab 
- [x] on first page, put disclaimer before next button 
- [x] income too high (disable results tab and button)
- [x] change field width to be consistent 
- [x] other legal two lines 
- [x] change back/clear/estimate 
- [x] spacing between intro para and next buttons 
- [x] add disclaimer (test site warning) at top of page (header) 

Let me know where you land with your two outstanding items and if I need to do anything to help you move along!


JeremyKennedy:
I think ideally when income is too high, that dialog comes from the backend. I think you could use what I provide now in the summary box to start, but I should probably update what I provide you so it provides text specifically related to income (currently it says "ineligible see below").

KhalidAdan:
> I think ideally when income is too high, that dialog comes from the backend. I think you could use what I provide now in the summary box to start, but I should probably update what I provide you so it provides text specifically related to income (currently it says "ineligible see below").

Good catch! I just copied and pasted the old text, I'll make the change now!

JeremyKennedy:
- [x] The spacing between the intro paragraph and the next button (first page) can probably be reduced @KhalidAdan 
- [x] Update text in summary box when income too high, so that it says income is too high @JeremyKennedy 
- [x] May need to add a new summary text property, one for when on the questions tab, one for on the results tab. This is because when we say "see below for more info" that is confusing when on the question tab. To discuss? Edit: Can hold off for now.

KhalidAdan:
> * [ ]  The spacing between the intro paragraph and the next button (first page) can probably be reduced @KhalidAdan

We can chat about this in our meeting at 2, it's like this in the designs :/

JeremyKennedy:
Regarding those two outstanding tasks from my comment - I'm probably going to skip those if I can't get them done today. I can't just go and change text without doing it "the right way" so I gotta wait for the incoming text changes to be provided to me. I really want to merge in our outstanding branches, and text changes can always be done later.

components/Layout/NeedHelpList.tsx
```diff
@@ -1,19 +1,62 @@
...
+      <div className="p-8 bg-emphasis rounded mt-8 md:mt-0">
+        <h2 className="h2">{title}</h2>
+        <ul className="pl-5 list-disc">
+          <li>
+            <a
+              className="text-default-text underline"
+              href="https://www.canada.ca/en/employment-social-development/corporate/contact/oas.html"
+              target="_blank" rel="noreferrer"
+            >
+              Contact Service Canada
```
JeremyKennedy: This link will be coming from the backend, I've made the change already in another branch, and just cherry-picked it into this one. So you can remove the SC link from your code and it'll come along with the rest of the links.

However, I noticed something a bit weird when testing this, on initial load, there is no request to the backend, so you don't use my links until there is user input. This is also causing an issue where, when you switch to Results immediately after first page load, you see this:
image

styles/globals.css
```diff
@@ -117,4 +117,11 @@
   .summary-link a {
     @apply text-default-text underline;
   }
+  .is-sticky {
```
JeremyKennedy, line 120: I am not sure why you made it wider when it starts floating, this causes overlap with the questions. See the dimensions in my screenshot to reproduce (or just resize and you should be able to see):
image

pages/eligibility/index.tsx
```diff
@@ -71,8 +58,9 @@ const Eligibility: NextPage<ResponseSuccess | ResponseError> = (props) => {
             className={({ selected }) =>
               selected
                 ? 'bg-white font-semibold p-2.5 pt-1.5 border border-t-4 border-content/90 border-r-muted/20 border-b-muted/20  border-l-muted/20 mr-2'
-                : 'bg-[#EBF2FC] font-semibold p-2.5 border border-muted/20 disabled mr-2'
+                : 'bg-[#EBF2FC] font-semibold p-2.5 border border-muted/20 disabled mr-2 disabled:cursor-not-allowed disabled:bg-muted disabled:text-white'
```
JeremyKennedy: The disabled Results/Estimate coloring is too dark imo, a lighter touch would appear more "disabled" to me. But, if you and Alla are cool with the color, then that's fine with me. It just looks dark, and not disabled, to me.

styles/globals.css
JeremyKennedy: It is also too far to the right when the page is wide. I think it should stay where it is, with the size it has, and just shift down, if possible.

JeremyKennedy (changes requested):
- Changing page width between 991-992 increases the text size of the TEST SITE header, this seems unintentional as no other text size changes. Edit: this is hit-and-miss, this is probably related to the finicky issue we had earlier with Edge's responsive resizing. May be able to ignore this one, then. Edit2: Hmm, when I resize slowly, the text size changes, when I resize quickly, it stays. Maybe hitting a specific width is triggering something, so when I go fast it doesn't hit that specific width? Yeah... when I use the dimensions at the top rather than dragging, it's consistent.... mostly...
- Related to above, when page width is specifically 992, the progress bar is messed up:
image

To-do:

- [x] Ping the backend on page load to properly populate links and the Results More Info box? Or is this related to SSE and you maybe need to update the defaults? Not sure, I saw a network call to eligibility.json that appears to return the good stuff, so idk why it's not visible: 
image
- [x] Assuming the above is possible, remove the static contact SC link from your end
- [x] Unless the current design is approved with Alla, I'd lighten up the disabled elements
- [x] Make the Need Help floating box work more nicely on extra wide or extra narrow screens. Also, it's weird at width=992. It's fine with width=993. When width=992, it gets funky... it seems to not move at all? When width=991, it moves down to the bottom as intended for mobile. When width is really wide, like over 2000, it's too far to the right. This is most apparent on a 27" screen at 1440p (meaning this isn't some silly specific screen size like the 992 issues above).

components/Layout/NeedHelpList.tsx
KhalidAdan: There should be a request to the backend, it just wont show up in the browser because it's executing in node. I think the summary object is not being set when the form is being set, should be a quick fix!

pages/eligibility/index.tsx
KhalidAdan: I think there might be a disabled style I can yoink, but it might fail accessibility requirements the way it is. I can change the text colour to black and update the grey we use. maybe? I'll let Alla handle the final decisions.

styles/globals.css
KhalidAdan: Keeping in the same spot will most likely cover up the same question you screen grabbed above, but I don't mind it too much as users can scroll up to see the questions. I can go ahead and try this change.

KhalidAdan:
> * Changing page width between 991-992 increases the text size of the TEST SITE header, this seems unintentional as no other text size changes. Edit: this is hit-and-miss, this is probably related to the finicky issue we had earlier with Edge's responsive resizing. May be able to ignore this one, then. Edit2: Hmm, when I resize slowly, the text size changes, when I resize quickly, it stays. Maybe hitting a specific width is triggering something, so when I go fast it doesn't hit that specific width? Yeah... when I use the dimensions at the top rather than dragging, it's consistent.... mostly...

This was actually code directly from SClabs, but I've removed the code for resizing text for consistency. If they want it back it's a very quick change.

> * Related to above, when page width is specifically 992, the progress bar is messed up:
>   image

The `useMediaQuery` hook had a bug! It is now fixed but meant that 992 px was always a problem width for us 🤦🏾 

> To-do:
> 
> * [x]  Ping the backend on page load to properly populate links and the Results More Info box? Or is this related to SSE and you maybe need to update the defaults? Not sure, I saw a network call to eligibility.json that appears to return the good stuff, so idk why it's not visible:

See comment below, fixed in `ComponentFactory,tsx`

> * [x]  Assuming the above is possible, remove the static contact SC link from your end
> * [x]  Unless the current design is approved with Alla, I'd lighten up the disabled elements
> * [x]  Make the Need Help floating box work more nicely on extra wide or extra narrow screens. Also, it's weird at width=992. It's fine with width=993. When width=992, it gets funky... it seems to not move at all? When width=991, it moves down to the bottom as intended for mobile. When width is really wide, like over 2000, it's too far to the right. This is most apparent on a 27" screen at 1440p (meaning this isn't some silly specific screen size like the 992 issues above).

All should be addressed!

components/Layout/NeedHelpList.tsx
JeremyKennedy: Confirmed fixed!

styles/globals.css
JeremyKennedy: Confirmed fixed!

pages/eligibility/index.tsx
JeremyKennedy: Yeah, it's a bit better I think. We'll see what the team says.

JeremyKennedy (approved):
All the fixes are great! Last thing is to make Need Help full width on mobile, but I'll pre-approve anyway!

PR #117, opened 2022-01-25
Rework frontend tooltip definition implementation

JeremyKennedy:
I took the liberty of rewriting the tooltip handling to use a similar translation approach as the backend, as well as adding support to reference other tooltip text values. The latter is not visible in this PR, but will be used by the New Partner Questions PR #111. You can actually check the deployment for that PR to see how this PR affects that one.

It seems to work as expected, I don't think I broke anything. Obviously, as proper language support isn't in the frontend, I default to English, and the French file is still all just English.


KhalidAdan (approved):
This looks good! Simple and clean is always the move, great job Jeremy! Messed around with the site and it all looks great!

PR #121, opened 2022-01-27
Proposed change to form layout 

KhalidAdan:
Proposing making the below that would make all question types the same width and prevent overlapping of the needhelp component while scrolling. A one liner!

**Before** (as in mockups):

https://user-images.githubusercontent.com/18726938/151391159-bbdc071f-487e-4a5b-b6f2-5ef878787923.mp4

**After**:

https://user-images.githubusercontent.com/18726938/151391175-b55f570e-50a2-4cc8-bb03-4584fbb23857.mp4


KhalidAdan:
WIP -- need to rebase develop and fix conflicts, just need the go ahead from the team

**Edit**: Spoke with Alla and she is not a fan of the change, this might sit until our Monday meeting where we can talk as a team

**Edit 2**: This is needed to fix a bug for demos, so it need's to go out. If we end up removing things, we can do so in a followup PR!

JeremyKennedy (approved):
Looks great to Wiam, Joan, and myself, lol. Approved!

PR #122, opened 2022-01-27
Added smooth scroll to the results page when estimate is clicked

KhalidAdan:
### Description
Adds a smooth scroll to the results page when estimate is clicked

Video:

https://user-images.githubusercontent.com/18726938/151420166-1b5323a0-0f41-4517-b635-367caac5bf20.mp4

**Note**: This will only smooth scroll if the alert is not in the viewport already, so the user can switch between tabs easily

### DIFF
--- a/components/Alert/index.tsx
+++ b/components/Alert/index.tsx
@@ -6,11 +6,12 @@ import { useWindowWidth } from '../Hooks'
 const OFFSET_WIDTH = 28
 
 export const Alert: React.VFC<{
+  id?: string
   title: string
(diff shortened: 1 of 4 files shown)


JeremyKennedy (changes requested):
- [x] On mobile especially, going from intro to questions results in scrolling to the top of the test banner, which is a bit unexpected. ~Ideally it would scroll to the app title.~
- [x] On mobile especially, going from questions/results to intro is same as above.

JeremyKennedy (approved):

PR #123, opened 2022-01-27
Add support for specifying location of links

JeremyKennedy:
WIP.

Let's us specify the link location. This is so that we can display CPP link only on the results page, and so we can display "apply here" link above the current links on the results page.

You would probably add fancy styling when location is RESULTS_APPLY, like that apply button we were talking about in the team meeting.

To-do:

- [x] Backend support
- [x] Frontend support


utils/api/definitions/enums.ts
```diff
@@ -78,3 +78,9 @@ export enum EstimationSummaryState {
   UNAVAILABLE = 'UNAVAILABLE', // yellow, can not provide any results, contact Service Canada (conditionally eligible)
   AVAILABLE_INELIGIBLE = 'AVAILABLE_INELIGIBLE', // red, display results (ineligible)
 }
+
+export enum LinkLocation {
+  STANDARD = 'STANDARD', // Questions > NeedHelp, and Results > MoreInfo
+  RESULTS_ONLY = 'RESULTS_ONLY', // Results > MoreInfo only
+  RESULTS_APPLY = 'RESULTS_APPLY', // new section above Results > MoreInfo
+}
```
KhalidAdan, line 87: Does this mean the more info link is the union of `STANDARD` and `RESULTS_ONLY`

JeremyKennedy: Hmm, I think so? I actually need to add another option, I'll do that now and maybe that'll clear things up?

KhalidAdan: So all this needs now is to have a link that is marked as RESULTS_APPLY so the results page will work and the needsHelpLink view to be updated. This is the link we use: https://www.canada.ca/en/services/benefits/publicpensions/cpp/old-age-security/apply.html

EDIT: As of now, if there is no link in the `links` object marked as RESULTS_APPLY the whole Next steps section is not rendered

JeremyKennedy:
Upon entering 65 into the age field, I get this:
image
This only happens after my latest commit, which added RESULTS_APPLY and RESULTS_ONLY links.

Edit: ahh, yeah, I also added a QUESTIONS_ONLY link, which is a new type I added today.

Edit: actually I think this is my bad! I removed links in one place and not in the other. fix incoming!

JeremyKennedy:
Ha! It works now! And it's great!

KhalidAdan (approved):
🚢

PR #124, opened 2022-01-27
Added a fix for tooltips being underneath the needhelp div and set the width to the correct width from the designs!

KhalidAdan:
As the title says!


JeremyKennedy (approved):
Both fixes look great! Thanks!

PR #126, opened 2022-01-28
Updated footer and added lang attribute to html tag

KhalidAdan:
Adds some padding and finalized the footer to core GC design specs. Also adds a `lang` attribute to the html tag!

This encompasses the work outlined in [SAEB-1108](https://jira-dev.bdm-dev.dts-stn.com/browse/SAEB-1108) and adds localized paths for languages (not full translation implementation)!

### DIFF
--- a/components/Layout/Footer.tsx
+++ b/components/Layout/Footer.tsx
@@ -69,7 +69,6 @@ export const Footer = () => (
         <li>
           <a href="#">Social media</a>
         </li>
-        #284162
         <li>
           <a href="#">Mobile applications</a>
         </li>
--- a/components/Layout/Footer.tsx
+++ b/components/Layout/Footer.tsx
@@ -1,61 +1,62 @@
 import Image from 'next/image'
+import Link from 'next/link'
 import styles from './Footer.module.css'
 
 export const Footer = () => (
   <footer className="text-[14px]">
     <div
-      className={`w-full h-auto bg-custom-blue-dark px-4 py-8 ${styles.footerBackground}`}
+      className={`w-full h-auto px-4 py-8 bg-custom-blue-dark ${styles.footerBackground}`}
     >
       <div
-        className="py-7 sm:container mx-auto"
+        className="sm:container mx-auto"
         role="navigation"
         aria-labelledby="footerNav1"
       >
         <h3 className="sr-only" id="footerNav1">
           About government
         </h3>
-        <ul className="flex flex-col text-xs lg:grid lg:grid-cols-2 xl:grid xl:grid-cols-3 lg:gap-1">
-          <li className="text-white w-64 md:w-56 lg:w-80 my-2.5 hover:underline">
+        <ul className="flex flex-col text-xs lg:grid lg:grid-cols-3 lg:gap-1">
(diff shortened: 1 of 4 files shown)


next.config.js
```diff
@@ -1,3 +1,8 @@
 module.exports = {
+  i18n: {
+    locales: ['en', 'fr'],
+    defaultLocale: 'en',
+    localDetection: true,
```
JeremyKennedy: should this be `localDetection` or `localeDetection`?

components/Layout/Footer.tsx
```diff
@@ -63,39 +63,64 @@ export const Footer = () => (
...
+          width="200px"
+          height="40px"
+          src="/wmms-blk.svg"
+          alt="Symbol of the Government of Canada"
+        />
+      </div>
+    </div>
+    <div className="flex justify-between">
+      <div className="flex items-center md:hidden py-8 md:pt-0">
+        <p className="px-4">Top of page</p>
```
JeremyKennedy: This says "top of page" but I don't see that anywhere in the page, it this text supposed to be visible somewhere?

JeremyKennedy (changes requested):
- [x] The spacing in the footer is too high (screenshots sent in Teams)

components/Layout/Footer.tsx
KhalidAdan: Visible on mobile only, as in the GC design mockups. That's what the `md:hidden` is doing

components/Layout/Footer.tsx
```diff
@@ -1,101 +1,128 @@
...
-            src="/wmms-blk.svg"
-            alt="Symbol of the Government of Canada"
-          />
-        </div>
+    <div className="flex justify-between px-4 py-8 sm:container mx-auto">
+      <ul className="flex flex-col md:flex-row md:space-x-6 list-disc px-4 text-[#284162]">
+        <li>
+          <a href="#">Social media</a>
+        </li>
+        #284162
```
JeremyKennedy: image
This doesn't quite seem right! 😉

JeremyKennedy (changes requested):
Looks better, except for the one issue with "social media" at the bottom!

components/Layout/Footer.tsx
JeremyKennedy: It's better, but you should still remove the bullet point before Social Media, see this screenshot from canada.ca:
image

JeremyKennedy (changes requested):
Still need to remove bullet point before Social Media

components/Layout/Footer.tsx
JeremyKennedy: Also, not that it's a problem, but canada.ca has an off-white color, unlike ours which is just plain white.

KhalidAdan:
> Still need to remove bullet point before Social Media

Yikes, I missed all these, my fault! I completelyyy misunderstood and blame sleep deprivation!

JeremyKennedy (approved):
> Yikes, I missed all these, my fault! I completelyyy misunderstood and blame sleep deprivation!

Whoops! I thought I mentioned the bullet point, turns out I didn't so definitely my bad there! 😃 

Looks good! And I checked the mobile layout to find the Top of page button. 👍

PR #127, opened 2022-01-28
Fixed bug with aria labels on selects

KhalidAdan:
### Description

Fixes a bug where the name field was not passed to the `FormSelect` component, which broke accessiblity.


JeremyKennedy (approved):
Appears to have zero impact on UI/UX (obviously except for accessibility purposes), so all good here!

PR #129, opened 2022-01-28
Fixed a bug with headless ui tabs and responding to data and update the app title

KhalidAdan:
Annoyingly, headless ui (from the people who made out design system, tailwind css) created these tabs in a way where re-rendering is pretty limited and shouldn't be relied on. Essentially, all the rendering work is only done reliably when there is a change event on the main tab component.

To fix this, I take control of the dom from react and manually set the disabled state as an effect of the currency field changing!

### DIFF
--- a/components/Forms/ComponentFactory.tsx
+++ b/components/Forms/ComponentFactory.tsx
@@ -47,8 +47,20 @@ export const ComponentFactory: React.VFC<FactoryProps> = observer(
       root.setSummary(data.summary)
     }
 
-    // check if income is too high to participate in calculation
+    // check if income is too high to participate in calculation, and fix a bug with headless ui tabs where they only re-render on interaction
     const incomeTooHigh = form.validateIncome()
+    useEffect(() => {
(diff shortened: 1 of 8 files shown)


components/Forms/ComponentFactory.tsx
```diff
@@ -47,8 +47,20 @@ export const ComponentFactory: React.VFC<FactoryProps> = observer(
...
+        const results = document.getElementsByClassName('results-tab')[0]
+        if (results) {
+          if (form.isIncomeTooHigh) {
+            results.setAttribute('disabled', 'disabled')
+          } else {
+            results.removeAttribute('disabled')
+          }
+        }
+      }
+    }, [incomeTooHigh])
```
KhalidAdan, line 63: real guts of the change here, useEffect is ran every time incomeTooHigh changes

JeremyKennedy (approved):
Looks good, title is fixed, autocomplete is disabled, the results tab is disabled when expected.

PR #132, opened 2022-01-31
Just a quick PR to attempt and solve the build warning vercel is giving us!

KhalidAdan:


JeremyKennedy (approved):
Confirming that the below errors are not visible anymore.
image

PR #133, opened 2022-01-31
Bugfix/focus visible bugfix

KhalidAdan:
### Description

This is a difficult one to test, so I'm not 1000% it's fixed but I'm pretty confident I got a useable solution here. The reason why I'm confident is I went down the tree and toggled all states until I found the div's that were problematic, and it looks good now!

https://user-images.githubusercontent.com/18726938/151829190-a03427bf-99d8-4b28-91e9-57dce9774eb8.mp4

### DIFF
--- a/components/Forms/ComponentFactory.tsx
+++ b/components/Forms/ComponentFactory.tsx
@@ -86,7 +86,7 @@ export const ComponentFactory: React.VFC<FactoryProps> = observer(
             ]}
           />
         )}
-        <div className="grid grid-cols-1 md:grid-cols-3 md:gap-10 mt-10 focus-visible:outline-none">
+        <div className="grid grid-cols-1 md:grid-cols-3 md:gap-10 mt-10">
           <form
             name="ee-form"
             data-testid="ee-form"
--- a/components/Forms/ComponentFactory.tsx
+++ b/components/Forms/ComponentFactory.tsx
@@ -60,7 +60,7 @@ export const ComponentFactory: React.VFC<FactoryProps> = observer(
           }
         }
       }
-    }, [incomeTooHigh])
+    }, [form.isIncomeTooHigh])
 
     return (
       <>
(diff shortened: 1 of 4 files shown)


JeremyKennedy (commented):
Seems the same before and after? 

https://user-images.githubusercontent.com/3630698/151900672-c61cea33-201f-4f35-b738-6755bda7bdce.mp4

Also, this PR contains unrelated changes from another PR, we should either remove those changes, or ensure the other PR is merged first and then squash this one.

KhalidAdan:
This is super odd, I can't get this to repro -- gonna throw some `!important`'s around and see if I can fix it that way.

Edit: This seems to show up when opening the dev tools only now, I wonder if this is just another Edge quirk. Investigating!

Edit 2: So, some unused (I thought) css was actually the fix that came with the template! 🤦🏾

JeremyKennedy (commented):
Is this the intended result? Not gonna lie, I don't really get the before/after here!


https://user-images.githubusercontent.com/3630698/152063102-a7237c6e-aeb5-4905-a57a-a217e96cc179.mp4

KhalidAdan:
> Is this the intended result? Not gonna lie, I don't really get the before/after here!
> 
>  msedge_J1l4KsvTCv.mp4

@JeremyKennedy 

So what you are doing there is forcing the element state, which is edge editing css with a hammer. What I did was make it so a user, without a hammer 😆, can't get this behaviour to occur. I was just using that tool to prove it really was happening!

It's pretty hard to test but what you can do is spam the dev tools while clicking around and you should never see this occur again. Spamming the dev tools forces the focus to move around, and when focusedElement == null a click on an element that can be focused will be given higher priority that usual.

_Hopefully,_ this is kind of a tricky one to ensure a fix for.

JeremyKennedy (approved):
I can't really test this, but you're just uncommenting stuff that came from the client, and you've played with it a bunch, so chances are it's all good.

Approved!

PR #140, opened 2022-02-01
Summary links now persist across clears

KhalidAdan:
Simple fix, just sends a new request. This introduces a light flicker, but is not egregious.


JeremyKennedy (approved):
It flickers a bit, but it works!

PR #141, opened 2022-02-01
ALW now returns "unavailable" when partner has partial OAS

JeremyKennedy:
We should not have been providing an entitlement amount for ALW when their partner has partial OAS. This is confirmed by the following from the government site for the entitlement table we are using: 
> Table 4: If you are receiving a full Old Age Security pension and your spouse or common-law partner is aged 60 to 64

This PR fixes that by providing no entitlement estimate when the partner has partial OAS.


KhalidAdan (approved):
👍🏾

PR #143, opened 2022-02-02
French translation update

KhalidAdan:
# [SAEB-1086](https://jira-dev.bdm-dev.dts-stn.com/browse/SAEB-1086)

This adds translation into the frontend portion of the EE. You can look at specific commits to see all that was added (there are a lot, and will probably be squashed) but the main points are:

1. Adds the `i18n/web` files in the same approach as the `api` package
2. Adds a new `useTranslation` hook that will query the files
3. Removes the `LanguageProvider` and relies solely on NextJS to handle languages/locales

Locales are ties to the url, and can be read more about [here](https://nextjs.org/docs/advanced-features/i18n-routing)


components/Hooks/index.tsx
```diff
@@ -102,3 +89,13 @@ export const useWindowWidth = () => {
...
+
+export const useInternationalization = (key: string): string => {
+  const { locale } = useRouter()
+  return webDictionary[locale][key]
+}
+
+export const useTranslation = (): WebTranslations => {
+  const { locale } = useRouter()
+  return webDictionary[locale]
+}
```
JeremyKennedy, line 98: `useTranslation` and `useInternationalization` seem like synonyms, but they are implemented differently. I think you should rename these to `getTranslations` and `getTranslationByKey`, respectively. And to clean things up, you could reference `getTranslations` within `getTranslationByKey`, as they share some logic.

components/Layout/Footer.tsx
```diff
@@ -1,128 +1,132 @@
...
-            <a className="font-body" href="#">
-              Treaties, laws and regulations
-            </a>
+        <div
+          className="sm:container mx-auto"
+          role="navigation"
+          aria-labelledby="footerNav1"
+        >
+          <h3 className="sr-only" id="footerNav1">
+            About government
```
JeremyKennedy: to-do?

components/Layout/Footer.tsx
```diff
@@ -1,128 +1,132 @@
...
+          <li>
+            <a href="#">{tsln.socialLink5}</a>
           </li>
         </ul>
+        <div className="hidden md:block">
+          <Image
+            width="200px"
+            height="40px"
+            src="/wmms-blk.svg"
+            alt="Symbol of the Government of Canada"
```
JeremyKennedy: to do? (it is just alt text, so it couuullldd slide)

components/Layout/Footer.tsx
```diff
@@ -1,128 +1,132 @@
...
+              </span>
+            </a>
+          </Link>
+        </div>
+        <div className="flex items-center px-4 md:hidden">
+          <Image
+            width="135px"
+            height="45px"
+            src="/wmms-blk.svg"
+            alt="Symbol of the Government of Canada"
```
JeremyKennedy: same - to do, alt text

i18n/web/en.ts
```diff
@@ -0,0 +1,79 @@
+import { WebTranslations } from '.'
+
+const en: WebTranslations = {
```
JeremyKennedy, line 3: 👌

i18n/web/en.ts
```diff
@@ -0,0 +1,79 @@
...
+  next: 'Next',
+  questions: 'Questions',
+  results: 'Results',
+  needHelp: 'Need help',
+  faq: 'Frequently asked questions (FAQ)',
+  estimate: 'Estimate',
+  required: 'required',
+  homePageP1:
+    'Canadian Old Age Benefits Estimator is a prototype in development. Based on the information you provide, this will estimate your eligibility for the Old Age Security (OAS) and Guaranteed Income Supplement (GIS). If eligible to receive the benefit, it will also estimate your monthly payment.',
+  homePageP2: 'The estimator covers 4 benefits programs:',
```
JeremyKennedy: imo: change "4" to "four"

i18n/web/index.ts
```diff
@@ -0,0 +1,69 @@
+import en from './en'
+import fr from './fr'
+
+export const webDictionary = { en, fr }
+
+export type WebTranslations = {
```
JeremyKennedy, line 6: beautiful!

pages/index.tsx
```diff
@@ -2,75 +2,41 @@ import type { GetStaticProps, NextPage } from 'next'
 import Link from 'next/link'
 import { useRouter } from 'next/router'
 import { useState } from 'react'
+import ts from 'typescript'
```
JeremyKennedy: Seems this import was added, but is unused (a few are unused actually):

image

JeremyKennedy (commented):
Looking really good so far! Just reviewed the code, didn't play with the app yet.

components/Layout/Footer.tsx
KhalidAdan: yeah wasn't in the doc, will fix!

pages/index.tsx
KhalidAdan: I'm gonna set my vscode to just do this for me, no more manually fixing these!

components/Hooks/index.tsx
KhalidAdan: This is a really good observation! For best practice reasons, hooks should be called `useX` but I think I'll just merge these two and make it semi generic!

i18n/web/en.ts
JeremyKennedy: marked resolved with no comment? I just asked in Slack about this to follow-up.

i18n/web/en.ts
```diff
@@ -0,0 +1,79 @@
...
+  socialLink4: 'Terms and conditions',
+  socialLink5: 'Privacy',
+  privacyDisclaimer:
+    'The personal information provided is governed in accordance with the Privacy Act. The estimator does not collect information that would enable personal identification. Your anonymous results may be collected for research purposes. ',
+
+  category: {
+    incomeDetails: 'Income Details',
+    personalInformation: 'Personal Information',
+    legalStatus: 'Legal Status',
+  },
```
JeremyKennedy, line 66: You deleted categories that we still use (partner details and social countries), how come? And as per the updated spreadsheet, the second word in these should not be capitalized.

JeremyKennedy: oh, I think the category deletion is because you don't use those specific ones for the progress bar, and rely on my response for these categories?

KhalidAdan: Yup!

KhalidAdan:
@JeremyKennedy any tips on mocking fetch? I disabled a test to get the green checkmark but Ideally this would just work.

JeremyKennedy:
> any tips on mocking fetch? I disabled a test to get the green checkmark but Ideally this would just work.

did you figure this out? not sure I 100% understood the question, and you pushed a bunch of commits after, so not sure if you got it or not?

Looking at `mst.test.ts`, I see you commented out / deleted some code that was mocking an API response, I'm assuming that's what you mean, but you also import `mockPartialGetRequest` which is what I'd recommend for this purpose... so I'm not sure! Let me know if you need anything.

KhalidAdan:
> > any tips on mocking fetch? I disabled a test to get the green checkmark but Ideally this would just work.
> 
> did you figure this out? not sure I 100% understood the question, and you pushed a bunch of commits after, so not sure if you got it or not?
> 
> Looking at `mst.test.ts`, I see you commented out / deleted some code that was mocking an API response, I'm assuming that's what you mean, but you also import `mockPartialGetRequest` which is what I'd recommend for this purpose... so I'm not sure! Let me know if you need anything.

I did end up figuring it out! `mockPartialGetRequest` is what I'm using all over the place and it is a lifesaver!

client-state/models/Form.ts
```diff
@@ -180,6 +187,7 @@ export const Form = types
           field.setError(d.message)
         }
       } else {
+        console.log(data)
```
JeremyKennedy: to remove?

components/Forms/ComponentFactory.tsx
```diff
@@ -172,8 +178,14 @@ export const ComponentFactory: React.VFC<FactoryProps> = observer(
...
                             ? [
-                                { key: 'true', text: 'Yes' },
-                                { key: 'false', text: 'No' },
+                                {
+                                  key: 'true',
+                                  text: router.locale == 'en' ? 'Yes' : 'Oui',
+                                },
+                                {
+                                  key: 'false',
+                                  text: router.locale == 'en' ? 'No' : 'Non',
```
JeremyKennedy: shouldn't yes/no/oui/non be in the translations file as well?

components/Forms/ComponentFactory.tsx
```diff
@@ -91,7 +92,12 @@ export const ComponentFactory: React.VFC<FactoryProps> = observer(
             className="col-span-2"
             noValidate
           >
-            <input type="hidden" name="_language" value={'EN'} />
+            <input
+              type="hidden"
+              name="_language"
+              id="_language"
+              value={router.locale == 'en' ? Language.EN : Language.FR}
```
JeremyKennedy, line 99: In the backend, Locale is like `en-CA` and `fr-CA`, while Language is `EN` or `FR`.  How come you are checking for a locale of `en`? Shouldn't it be `en-CA`? If not, then why not use Language instead of Locale, so you don't need these sorts of conversions?

JeremyKennedy: (this comment applies elsewhere as well)

components/Forms/ComponentFactory.tsx
```diff
@@ -209,24 +221,33 @@ export const ComponentFactory: React.VFC<FactoryProps> = observer(
...
-                    selectedTabIndex(1)
+                  if (
+                    !form.validateAgainstEmptyFields(router.locale) &&
+                    !form.hasErrors
+                  ) {
+                    const language = document.querySelector(
+                      '#_language'
+                    ) as HTMLInputElement
+                    root.setCurrentLang(
+                      language.value == 'EN' ? Language.EN : Language.FR
```
JeremyKennedy: isn't this redundant? can't it just be `language.value`? as it is always either EN or FR

pages/eligibility/index.tsx
```diff
@@ -31,7 +32,9 @@ const Eligibility: NextPage<ResponseSuccess | ResponseError> = (props) => {
 export const getStaticProps: GetStaticProps = async (context) => {
   // using mockPartialGetRequest() is simply a convenient way of calling
   // the backend function, as it expects specific request/response objects
-  const data = await mockPartialGetRequest({})
+  const data = await mockPartialGetRequest({
+    _language: context.locale == 'en' ? Language.EN : Language.FR,
```
JeremyKennedy, line 36: locale comment applies here too

components/Forms/ComponentFactory.tsx
KhalidAdan: 100%!

components/Forms/ComponentFactory.tsx
KhalidAdan: This was me juking the mst type. `string !== Language` but I can just use `as` instead.

components/Forms/ComponentFactory.tsx
KhalidAdan: I went with en and fr over those because they are used to generate the URL for the French side and the html lang attribute needs to be en or fr.

`/fr/eligibility` vs `/fr-CA/eligibility`

[SCLabs nextJS app using `fr` over `fr-CA`](https://alpha.service.canada.ca/fr/projets/centre-numerique)

Since this has to go out, I was thinking maybe I put up a follow-up using NextJS masking (atm not sure if it will even work) to maybe remove the `/fr/...` from the URL completely.

JeremyKennedy (changes requested):
This header needs to be translated:
image

JeremyKennedy:
As does this:
image

JeremyKennedy (approved):

PR #144, opened 2022-02-04
Adding compatibility for form to fill itself based on URL

KhalidAdan:
### Description

Adds the ability to copy a form field's data as a URL and parses incoming url's to fill out form data.

`income=20000&age=65&maritalStatus=married&livingCountry=CAN&legalStatus=canadianCitizen&canadaWholeLife=false&yearsInCanadaSince18=40&partnerBenefitStatus=none&partnerIncome=0`

### Additional Notes

Is it possible to ignore URL parameters that are not valid for the backend instead of failing? Currently if the user edits / copies a link and misses something, the form breaks completely. Currently, the frontend will just ignore incorrect parameters by checking if a field exists with the key exists, can we do something similar on the backend?


KhalidAdan:
This is on hold indefinitely at the moment, form data might be Protected A and as such cannot be transported through a URL

JeremyKennedy:
This PR has been updated to only include parsing the query string. The UI for copying the URL to the clipboard has been removed as per the above comment. ~Let's merge this in to support the CSV feature!~ Actually, I think I'll merge the CSV PR without this feature to start, and then I'll build off this branch in order to add this to the CSV data.

KhalidAdan:
@JeremyKennedy sure thing! Just delete this branch when it's no longer needed!

PR #145, opened 2022-02-04
There is a delay between clearing the form and sending a new request …

KhalidAdan:
There is a delay between clearing the form and sending a new request to fill it, so the helper link functions need to be emptied on clear and return null or an empty array

Currently, clearing breaks the app after you visit the results page

### DIFF
--- a/client-state/store.ts
+++ b/client-state/store.ts
@@ -1,10 +1,14 @@
-import { types, IMaybe, ISimpleType, ModelCreationType } from 'mobx-state-tree'
 import {
-  ExtractCFromProps,
   IArrayType,
+  IMaybe,
   IModelType,
+  Instance,
+  ISimpleType,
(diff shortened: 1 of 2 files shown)


client-state/store.ts
```diff
@@ -1,10 +1,14 @@
...
+  IMaybe,
   IModelType,
+  Instance,
+  ISimpleType,
+  ModelCreationType,
+  types,
   _NotCustomized,
-} from 'mobx-state-tree/dist/internal'
+} from 'mobx-state-tree'
+import { ExtractCFromProps } from 'mobx-state-tree/dist/internal'
```
KhalidAdan, line 11: import less from internals

JeremyKennedy (approved):
Yep, it's not broken anymore!

PR #149, opened 2022-02-07
Backend Text Updates

JeremyKennedy:
I've pulled all updates from the spreadsheet. Once notable improvement is we can now use `{VARIABLES}` in the text, which will be replaced automatically. For example, if in the Excel sheet a text contains `{ENTITLEMENT_AMOUNT}`, there is a rule which will replace this with the true entitlement amount. 

I will note that the returned summary object now contains the sum of entitlements. Not sure if this will break types on your end. You may optionally use this for the sum in the table, rather than doing the math on your end.

I also added a LinkLocation of `HIDDEN` which should not show anywhere and is just used for linkifying strings, like `{LINK_SERVICE_CANADA}`.

This PR also simplifies/unifies currency formatting, by adding `numberToStringCurrency()` to format currencies. This should be used in the frontend too, for consistency and easy updating. I've asked Wiam what formatting we want to standardize on, currently I'm using ISO standard which is `1 000,00 $` for French, which might be weird.

A note and link on OAS deferral will be visible when age is 64-70 and OAS eligible.

### DIFF
--- a/i18n/api/index.ts
+++ b/i18n/api/index.ts
@@ -1,13 +1,9 @@
+import { Language, Locale } from '../../utils/api/definitions/enums'
 import { Link } from '../../utils/api/definitions/types'
 import en from './en'
 import fr from './fr'
 
 const apiTranslationsDict = { en, fr }
-
-export enum Language {
-  EN = 'EN',
-  FR = 'FR',
-}
 
 export interface KeyAndText {
   key: string
@@ -15,6 +11,8 @@ export interface KeyAndText {
 }
 
 export interface Translations {
+  _language: Language
+  _locale: Locale
   benefit: { oas: string; gis: string; allowance: string; afs: string }
   category: {
     incomeDetails: string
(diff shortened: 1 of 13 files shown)


KhalidAdan (approved):
Looks solid!

i18n/api/index.ts
```diff
@@ -128,3 +130,21 @@ export function getTranslations(language: Language): Translations {
...
+  number: number,
+  locale: Locale,
+  options?: { rounding?: number }
+): string {
+  const rounding = options?.rounding === undefined ? 2 : options.rounding
+  return number.toLocaleString(locale, {
+    style: 'currency',
+    currency: 'CAD',
+    currencyDisplay: 'narrowSymbol',
+    minimumFractionDigits: rounding,
```
KhalidAdan, line 146: What a cool function, adding to my lexicon!

KhalidAdan (approved):
:ship:

PR #150, opened 2022-02-07
Consistent "end of entitlement checks" handling, remove unused reasoning text

JeremyKennedy:
Small cleanup. Crossing my fingers that the removal of that last `this.input.livingCountry.noAgreement` condition doesn't break anything, but as far as I could tell with the way the conditions are now, it was impossible to reach that code anyways.

There should never be a case where that `entitlement logic failed to produce a result` error is reached, if there is I need to add in more conditions.

Also removed unneeded GIS Apply link.

### DIFF
--- a/i18n/api/en.ts
+++ b/i18n/api/en.ts
@@ -376,12 +376,6 @@ const en: Translations = {
       order: 5,
       location: LinkLocation.RESULTS_APPLY,
     },
-    gisApply: {
-      text: 'Guaranteed Income Supplement: Application',
-      url: 'https://www.canada.ca/en/services/benefits/publicpensions/cpp/old-age-security/guaranteed-income-supplement/apply.html',
-      order: 6,
-      location: LinkLocation.RESULTS_APPLY,
-    },
     alwApply: {
       text: 'Guaranteed Income Supplement - Allowance application',
       url: 'https://www.canada.ca/en/services/benefits/publicpensions/cpp/old-age-security/guaranteed-income-supplement/allowance/apply.html',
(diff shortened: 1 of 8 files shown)


i18n/api/en.ts
```diff
@@ -378,12 +376,6 @@ const en: Translations = {
       order: 5,
       location: LinkLocation.RESULTS_APPLY,
     },
-    gisApply: {
-      text: 'Guaranteed Income Supplement: Application',
-      url: 'https://www.canada.ca/en/services/benefits/publicpensions/cpp/old-age-security/guaranteed-income-supplement/apply.html',
-      order: 6,
-      location: LinkLocation.RESULTS_APPLY,
-    },
```
KhalidAdan, line 386: 👍🏾

utils/api/benefits/afsBenefit.ts
```diff
@@ -128,15 +128,8 @@ export class AfsBenefit extends BaseBenefit {
           detail: this.translations.detail.dependingOnLegal,
         }
       }
-    } else if (this.input.livingCountry.noAgreement) {
-      return {
-        result: ResultKey.INELIGIBLE,
-        reason: ResultReason.SOCIAL_AGREEMENT,
-        detail: this.translations.detail.ineligibleYearsOrCountry,
-      }
```
KhalidAdan, line 136: Devil's advocate here, should this condition stay? If we aren't sure it's unreachable is it safer to just leave it in? It's technically correct and a valid reason the calculator should spit out.

Just playing the other side, if you're confident -- we can try this and if it breaks we just fix it later

KhalidAdan (approved):
:shipit:

utils/api/benefits/afsBenefit.ts
JeremyKennedy: I'm pretty confident it's fine! I spent a little while going through the conditions and as far as I can tell it's totally unreachable. I wanted to get rid of it because the reason detail text is way too long. If it becomes an issue we can handle it better than it was being handled before this PR's changes.

PR #151, opened 2022-02-07
Link to additional reasons within table when client is ineligible

JeremyKennedy:
Now when the client is ineligible, the results detail text now includes a link to view the possible remaining ineligibility reasons.

To-do:

- [x] unsafe html in frontend for the detail text within the table (I just pushed a change doing this but the links are not blue as they should be) @KhalidAdan 
- [x] review this with the team and see if we like it

### DIFF
--- a/i18n/api/en.ts
+++ b/i18n/api/en.ts
@@ -345,6 +345,8 @@ const en: Translations = {
       'You may be eligible to receive this benefit when you turn 60, depending on your legal status in Canada. We encourage you to contact Service Canada for a better assessment.',
     dependingOnLegalWhen65:
       'You may be eligible to receive this benefit when you turn 65, depending on your legal status in Canada. We encourage you to contact Service Canada for a better assessment.',
+    additionalReasons:
+      '{LINK_MORE_REASONS} for possible additional ineligibility reasons.',
   },
   summaryTitle: {
     moreInfo: 'More information needed',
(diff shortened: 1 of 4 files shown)


i18n/api/en.ts
```diff
@@ -338,14 +358,15 @@ const en: Translations = {
   },
   summaryDetails: {
     moreInfo:
-      'Please fill out the form. Based on the information you provide, the application will estimate your eligibility. If you are a qualified candidate, the application will also provide an estimate for your monthly payment.',
+      'Please fill out the form. Based on the information you will provide today, the application will estimate your eligibility. If you are a qualified candidate, the application will also provide an estimate for your monthly payment.',
     unavailable:
-      'Based on the information provided, we are unable to determine your eligibility. We encourage you to contact Service Canada using the link below.</br><a href="https://www.canada.ca/en/employment-social-development/corporate/contact/oas.html" target="_blank">Contact Service Canada</a>',
+      'Based on the information you provided today, we are unable to determine your eligibility. We encourage you to contact {LINK_SERVICE_CANADA}.',
```
KhalidAdan, line 361: @JeremyKennedy  We'll need to add a class to the `<a>` tags (as tailwind makes all elements unstyled by default) that are templated here? Any ideas?

utils/api/helpers/requestHandler.ts
```diff
@@ -409,6 +424,22 @@ export class RequestHandler {
...
+        `<a href="${translations.links.gisReasons.url}" target="_blank">${translations.links.gisReasons.text}</a>`
+      )
+      .replace(
+        '{LINK_MORE_REASONS_ALW}',
+        `<a href="${translations.links.alwReasons.url}" target="_blank">${translations.links.alwReasons.text}</a>`
+      )
+      .replace(
+        '{LINK_MORE_REASONS_AFS}',
+        `<a href="${translations.links.afsReasons.url}" target="_blank">${translations.links.afsReasons.text}</a>`
+      )
```
KhalidAdan, line 442: 👍🏾

KhalidAdan (approved):

PR #155, opened 2022-02-08
Static Site Generation

JeremyKennedy:
Switch from "on request server side rendering" to "on build static site generation".

The only thing this broke is the unmerged "query string parsing to form" branch, which we dealt with!


KhalidAdan (approved):
E X C I T I N G !!!

PR #156, opened 2022-02-08
Bugfix/inconsistent progress rules 

KhalidAdan:
Fixing a bug where progress steps could be marked as complete if the step before it was marked as incomplete

### DIFF
--- a/client-state/models/Form.ts
+++ b/client-state/models/Form.ts
@@ -53,10 +53,10 @@ export const Form = types
       const iComplete =
         incomeFields.length > 0 && incomeFields.every((field) => field.filled)
       const pComplete =
-        personalFields.length > 0 &&
+        personalFields.length > 0 && iComplete &&
         personalFields.every((field) => field.filled)
       const lComplete =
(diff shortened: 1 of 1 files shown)


client-state/models/Form.ts
```diff
@@ -53,10 +53,10 @@ export const Form = types
       const iComplete =
         incomeFields.length > 0 && incomeFields.every((field) => field.filled)
       const pComplete =
-        personalFields.length > 0 &&
+        personalFields.length > 0 && iComplete &&
         personalFields.every((field) => field.filled)
       const lComplete =
-        legalFields.length > 0 && legalFields.every((field) => field.filled)
+        legalFields.length > 0 && pComplete && legalFields.every((field) => field.filled)
```
KhalidAdan, line 59: personal cannot be complete unless income is, and legal cannot be complete if the other two aren't

JeremyKennedy (approved):
It's a bit unfortunate that Partner Details questions are required to satisfy the Personal Information progress bar, but I don't see an easy way around that. Seems to work great otherwise.

PR #159, opened 2022-02-09
Bolded the benefit names on homepage

KhalidAdan:


JeremyKennedy:
I see you're bolding using a class rather than `<b>`, which made me check my "bold ENTITLEMENT_AMOUNT" commit, only to realize I didn't include the bold change as part of that commit... so, that brings me to my question, should I bold using the class as you've done here, is or `<b>` cool?

KhalidAdan:
> I see you're bolding using a class rather than `<b>`, which made me check my "bold ENTITLEMENT_AMOUNT" commit, only to realize I didn't include the bold change as part of that commit... so, that brings me to my question, should I bold using the class as you've done here, is or `<b>` cool?

In tailwind you'll probably want a class as tailwind has a really heavy handed CSS reset so there are no default styles applied. it used to be that we used bold, but if we wanted to be 100% semantically correct we should use `<strong class="font-bold">` if we want inline elements that are supposed to be bolded. 

Because of the CSS reset, the chosen element really doesn't matter in tailwind, as most people use a class on an element (I have taken to using `<span>`'s anytime I need a quick inline element in a `<p>` tag as *most* of the time you need bolding and some other styles).

JeremyKennedy (approved):
As per (my understanding of) your comment I used `strong` for the entitlement amount, Your change looks good to me!

KhalidAdan:
> As per (my understanding of) your comment I used `strong` for the entitlement amount, Your change looks good to me!

I've updated my original response for clarity. The CSS reset makes the `strong` element do nothing so you need to use my edit above.

PR #160, opened 2022-02-09
CSV Export

JeremyKennedy:
## PR 
This PR adds a new API endpoint, `exportCsv`, that takes the exact same request structure as the main endpoint, and returns a CSV file.

To-do:

- [x] Translate CSV file
- [ ] Add link to application including form inputs
- [x] Hook UI's CSV button to call the new endpoint (@KhalidAdan)
- [x] Provide proper French text

Here is what the CSV looks like in Excel. It is admittedly... not beautiful, due to my not following CSV standards. The Form Responses section has column A with the long text, the Estimation Results section has column C with the long text, and the links has column B with long text, making it pretty awkward to fit everything nicely. I have manually resized columns and applied word wrap to produce this screenshot showing the current state:

image

I would love to hear any suggestions you may have to make this more intuitive. I am wondering if it is worth supporting both Excel and CSV, but that would take some effort and we are starting to run low on time,

## Proposal (accepted)
I was thinking, given that we (hopefully only temporarily) are holding off on allowing the user to copy a link to their clipboard with the form responses, that we will want to capture the form responses in the CSV now.

I am considering not following the CSV spec (columns with values) and instead treating it like a table grid with multiple different columns/tables, like this:

image

An alternative is to provide an Excel file, with tabs for each of these tables. Could even provide the option of Excel or CSV.

Pros and cons of this "breaking spec with multiple tables in the csv file" approach:

Pros:
- user gets everything they would want in a single file, very valuable to see why they got the results they did
- user does not need excel installed to view, but if they do have it, it'll display nicely 

Cons:
- people who know how csv is supposed to be might be annoyed we aren't following that?

I had more in mind, but I forgot them for now, anyways... I think despite this CSV structure being "wrong", the value it provides outweigh that one disadvantage.


JeremyKennedy:
We could include the dynamic links in that links section, that could be cool? Or info overload... could have the apply now at least?

KhalidAdan:
I really like the proposal, and don't mind going against CSV spec. This is never going to be parsed back into a computer so we can bend it a little to make the humans happy! I just hope we don't get told to remove it because it is Protected A data. If a user shares their own data we can't be held responsible for that.

The apply link would be my choice over the dynamic form link.

Looks real good Jeremy!

KhalidAdan (approved):

KhalidAdan (approved):
🎩'd and everything seems like it works very well! A little bit of a bummer that CSV cannot be styled :(

LGTM!

PR #165, opened 2022-02-11
pushed tooltips changes

KhalidAdan:


JeremyKennedy (approved):
Looks good. Some tooltips are still English, but I can see it's because there is no text at all in the spreadsheet. I've highlighted those cells red and noted that we need something there.

PR #167, opened 2022-02-11
Fix broken sentence after results table

JeremyKennedy:
todo: 

- [x] fix the FAQ link

### DIFF
--- a/components/ContactCTA/index.tsx
+++ b/components/ContactCTA/index.tsx
@@ -1,13 +1,28 @@
 import Link from 'next/link'
+import { useLayoutEffect } from 'react'
 import { WebTranslations } from '../../i18n/web'
 import { useStore, useTranslation } from '../Hooks'
 
 export const ContactCTA: React.VFC = () => {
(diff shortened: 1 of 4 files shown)


components/ContactCTA/index.tsx
```diff
@@ -1,16 +1,31 @@
...
+  }
+
+  useLayoutEffect(() => {
+    const self = document.querySelectorAll('#faqLink a')[1]
+    if (!self) return
+
+    self.addEventListener('click', handleClick)
+    return () => self && self.removeEventListener('click', handleClick)
+  })
+
```
KhalidAdan, line 21: need to `useLayoutEffect` because we want to synchronously add the event listener after the DOM is available. One annoyance of `dangerouslySetInnerHTML` is that I can't access it via normal DOM functions, so I had to add an id on the paragraph and query down to the anchor tag to add the listener.

It works now though!

KhalidAdan (approved):

KhalidAdan (approved):

PR #173, opened 2022-02-14
Bugfix/tsln fixes

KhalidAdan:
Adds translation to `Select from X...` and the `moreInfo` header


JeremyKennedy (approved):
I made a few changes, it looks good!

PR #174, opened 2022-02-15
incorporate design feedback

KhalidAdan:
This PR adds all of the design fixes that were brought up as a result of Alla's meeting with the design stakeholders.

They are: 
- Add privacy disclaimer
- Update colours for some components
- Ensure estimated month total per benefit amount is a maximum of two rows
- Update required text after form questions to be translated correctly as obligatiore
- Right align estimates so entitlements are easier to read
- Remove bolding on form question labels
- New Tab design

This PR does not attempt to add the DECD components yet as they are npm only so that will come in a follow-up PR


.npmrc
```diff
@@ -0,0 +1 @@
+@dts-stn:registry=https://npm.pkg.github.com
```
KhalidAdan: remove

components/Forms/ComponentFactory.tsx
```diff
@@ -151,7 +158,7 @@ export const ComponentFactory: React.VFC<FactoryProps> = observer(
                         name={field.key}
                         label={field.label}
                         placeholder={field.placeholder ?? ''}
-                        onChange={field.handleChange}
+                        onChange={debounce(field.handleChange, 300)}
```
KhalidAdan, line 161: I went with a 300ms debounce as it felt right, but let me know what you think. This solves the issue where you want to enter 65 as an age but when you hit 6 it already has an error on it

JeremyKennedy (commented):
- I should have mentioned this earlier while I still had the chance, but I feel like the privacy notice should be merged into the original disclaimer. Oh well, this is fine too.
- Not in the changelog, but I see the Test Site header sizing is increased. Now, it is nearly the largest text on the screen, seems a little excessive now, and the alignment was better before:
image
vs after:
image
- Not in the changelog, but I see you redesigned the tabs to be what is apparently not tabs and what is somehow more accessible? I don't get how this resolves the issues people raised, but it saves us from a bigger rewrite, so let's take it!
- I can confirm the questions are not bolded anymore. I liked it better before, because it made the questions stand out from the radio buttons (the radio text stands out more than the question text now).
- I see the outline colors have been updated, but not consistently for all components:
image
- I see the Need Help background is more of a muted color
- The 300 debounce works for me. Could argue increasing it to say, 500, but I don't think anyone will complain.
- The new results table estimated amount changes look good to me.
- The changes to the Application button look good as well.

KhalidAdan:
Alla and I spoke and a couple of things were fixed that you brought up, plus label changes to make the page feel less uniform and have the higher importance elements stand out a little more!

JeremyKennedy (changes requested):
- :heavy_check_mark: I see the Test Site header is vertically aligned now, nice!
- :heavy_check_mark: I see the questions are back to bold, great!
- :question: I see the Progress Bar now has darker colors for the unchecked items. I think this makes it harder to visually differentiate between the checked and unchecked, because now the color is very similar. I think that should go back to before so the contrast between the two is higher.
Now:
image
Before:
image
- :exclamation: The outline on TextBox vs Dropdown is still inconsistent, before the latter was lighter, now it is darker:
image
- :question: I see the buttons have a heavier outline now, was this intended? It feels like it goes against the design system, but I don't have anything to back that up!
Now:
image
Before:
image

I think that is all the changes I've noticed since my last review, you didn't mention them specifically so I am unable to confirm specific changes.

I just have the two questions :question: above, then the inconsistent outline :exclamation:, then should be good to merge :heavy_check_mark:!

KhalidAdan:
> * ✔️ I see the Test Site header is vertically aligned now, nice!
> * ✔️ I see the questions are back to bold, great!
> * ❓ I see the Progress Bar now has darker colors for the unchecked items. I think this makes it harder to visually differentiate between the checked and unchecked, because now the color is very similar. I think that should go back to before so the contrast between the two is higher.
>   Now:
>   image
>   Before:
>   image

The design system team has commented that the colours we have do not pass accessiblity checks, so Alla implemented these new ones.

> * ❗ The outline on TextBox vs Dropdown is still inconsistent, before the latter was lighter, now it is darker:
>   image

This is a good observation, for some reason though -- the border is thicker on `react-select` components, but I can confirm they are indeed the same colour (#333). I'll address this now. Edit: A react select bug was assigning a heavier selector, they _were_ different colours!

> * ❓ I see the buttons have a heavier outline now, was this intended? It feels like it goes against the design system, but I don't have anything to back that up!
>   Now:
>   image
>   Before:
>   image
> 

Nope, just to heavy handed of a css variable and my browser extensions being too forgiving. Fixing now! I'll also start checking these in a private window with no extension.

> I think that is all the changes I've noticed since my last review, you didn't mention them specifically so I am unable to confirm specific changes.
> 
> I just have the two questions ❓ above, then the inconsistent outline ❗, then should be good to merge ✔️!

JeremyKennedy (approved):
The dropdown outline is fixed, the button outline is fixed, and I guess we're cool with the progress bar coloring being hard to differentiate visually, because it's more accessible somehow!

All good!

PR #176, opened 2022-02-15
Major refactor of RequestHandler

JeremyKennedy:
This is a refactor of the `RequestHandler` class. It should have no functional changes. Previously, I was focusing on using static functions, so it was obvious what was being read and written, and for easy reusability. Those pros have ultimately not outweighed the big con of adding unnecessary complexity, with passing around parameters all over the place as if I wasn't using a class at all. So, now the functions are not static, and will read variables from their class.


KhalidAdan (approved):
Tests pass and the change seems sensible enough, so I am confident that this should be an invisible change; LGTM!

PR #178, opened 2022-02-15
Do not show additional reasons main reason is client is too young

JeremyKennedy:
Previously, we were adding the "please see here for additional reasons" text when the client was ineligible, even if the only reason for ineligibility is age. This was unnecessary, as if the main reason is age then there are no other reasons to worry about. So, this PR will no longer add that text when the reason is they are too young.

### DIFF
--- a/utils/api/helpers/requestHandler.ts
+++ b/utils/api/helpers/requestHandler.ts
@@ -7,7 +7,11 @@ import { AfsBenefit } from '../benefits/afsBenefit'
 import { AlwBenefit } from '../benefits/alwBenefit'
 import { GisBenefit } from '../benefits/gisBenefit'
 import { OasBenefit } from '../benefits/oasBenefit'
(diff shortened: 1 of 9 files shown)


utils/api/helpers/requestHandler.ts
```diff
@@ -345,8 +365,9 @@ export class RequestHandler {
 
       // if client is ineligible, the table will be populated with a link to view more reasons
       const ineligibilityText =
-        result.eligibility.result === ResultKey.INELIGIBLE
-          ? ` ${translations.detail.additionalReasons}`
+        result.eligibility.result === ResultKey.INELIGIBLE &&
+        result.eligibility.reason !== ResultReason.AGE_YOUNG // do not add additional reasons when they will be eligible in the future
```
KhalidAdan, line 369: 👍🏾

KhalidAdan (approved):
🚢

PR #182, opened 2022-02-23
Addition of Adobe Analytics

KhalidAdan:
Added AA to the home page, and the push function is being called, but blocked on the AA site / checking functionality. How do we confirm the analytics is working? There seems to be one sample project with 11500 projects.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,3 +1,3 @@
 {
-  "extends": "next/core-web-vitals"
+  "extends": "next"
 }
--- a/client-state/store.ts
+++ b/client-state/store.ts
@@ -103,7 +103,7 @@ export const RootStore = types
       } else if (index == 2) {
         return 'faq'
       }
-      return '' // TODO: throw? Or unknown?
+      return 'unknown'
     },
   }))
   .actions((self) => ({
--- a/client-state/store.ts
+++ b/client-state/store.ts
@@ -94,6 +94,18 @@ export const RootStore = types
     activeTab: types.optional(types.number, 0),
     lang: types.enumeration(Object.values(Language)),
   })
+  .views((self) => ({
+    getTabNameForAnalytics(index: number) {
+      if (index == 0) {
+        return 'questions'
+      } else if (index == 1) {
+        return 'results'
+      } else if (index == 2) {
+        return 'faq'
+      }
+      return '' // TODO: throw? Or unknown?
+    },
+  }))
   .actions((self) => ({
     setActiveTab(num: number) {
       self.activeTab = num
--- a/components/Tabs/index.tsx
+++ b/components/Tabs/index.tsx
@@ -36,8 +36,8 @@ export const Tabs: React.FC<ResponseSuccess> = observer((props) => {
           win.adobeDataLayer.push({
             event: 'pageLoad',
             page: {
-              title: 'eng-sc-labs eligibility-estimator test',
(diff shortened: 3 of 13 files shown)


.eslintrc.json
```diff
@@ -1,3 +1,3 @@
 {
-  "extends": "next/core-web-vitals"
```
KhalidAdan, line 2: `core-web-vitals` tries to make us use the Script component instead of the browser supported synchronous script tag.

client-state/store.ts
```diff
@@ -94,6 +94,18 @@ export const RootStore = types
...
+  .views((self) => ({
+    getTabNameForAnalytics(index: number) {
+      if (index == 0) {
+        return 'questions'
+      } else if (index == 1) {
+        return 'results'
+      } else if (index == 2) {
+        return 'faq'
+      }
+      return '' // TODO: throw? Or unknown?
```
KhalidAdan: Leaning toward `''` being okay, if we can't tell which tab we're on we should still send a request to AA

KhalidAdan:
image

client-state/store.ts
KhalidAdan: going to go with unknown, better debugging breadcrumb

components/Tabs/index.tsx
```diff
@@ -12,13 +14,26 @@ import { ResultsPage } from '../ResultsPage'
...
 
   return (
     <Tab.Group
       key={root.activeTab}
       defaultIndex={root.activeTab}
       onChange={(index) => {
         root.setActiveTab(index)
+        if (process.browser) {
+          const win = window as Window &
+            typeof globalThis & { adobeDataLayer: any; _satellite: any }
```
JeremyKennedy, line 27: what the heck

pages/eligibility/index.tsx
```diff
@@ -1,15 +1,38 @@
...
+      const win = window as Window &
+        typeof globalThis & { adobeDataLayer: any; _satellite: any }
+      const lang = tsln.langLong
+      const creator = tsln.creator
+      const title =
+        lang +
+        '-sc labs-eligibility estimator-' +
+        root.getTabNameForAnalytics(0)
+
+      sendAnalyticsRequest(lang, title, creator, win)
```
JeremyKennedy, line 32: duplicated code from above, should be pulled into generic function

pages/index.tsx
```diff
@@ -1,68 +1,87 @@
...
+  useEffect(() => {
+    // only run on mount on the client
+    if (process.browser) {
+      const win = window as Window &
+        typeof globalThis & { adobeDataLayer: any; _satellite: any }
+      const lang = tsln.langLong
+      const creator = tsln.creator
+      const title = lang + '-sc labs-eligibility estimator-home'
+
+      sendAnalyticsRequest(lang, title, creator, win)
```
JeremyKennedy, line 25: appears to be similar duplicated code

JeremyKennedy (approved):
Good enough! Some duplicated code but it sounds like you want this out ASAP so I'll approve.

components/Tabs/index.tsx
KhalidAdan: Stuffing data into the window object and getting typescript to agree that I am correct! haha

KhalidAdan:
I'll make a note of the objections and swing back around to fix, ty for the speedy review!

PR #185, opened 2022-02-24
Entitlement when Partial OAS

JeremyKennedy:
I've completed the implementation of the logic for GIS entitlement calculations when the client has Partial OAS. I've also re- enabled ALW entitlement calculations when the client has Partial OAS.

Regarding the GIS calculations, I've implemented _my understanding_ of it. This could be wrong! However after reviewing with Wiam, we found some logic that appears to work and I've added it to our Data Patterns spreadsheet on the GIS Analysis tabs. The informative tab is also screenshotted below.

There is one piece of logic for when the client is under the maximum income limit. This is fairly straightforward, and I don't see there being any issues with this part. Since the publicly-available tables show a target Combined OAS+GIS amount, and this amount is accurate even when the client has Partial OAS, the math is simple here:
 
image

The other piece of logic is a bit more weird. Apparently, even if the client is above the maximum income limit for GIS, they may actually be entitled to GIS if they have Partial OAS. However, since the public tables do not detail the Combined OAS+GIS amount above the maximum income limit, I can't use the formula above unless I am able to extrapolate that. Luckily, I found a way that appears to work. Looking at the document provided to us which did have the Combined amounts over the max income, I saw that the Combined amount decreases by $1 for every income interval. Using this as the key, I came up with this logic:
image

And everything seems to work!

The weird bit is that one can be GIS eligible when above the GIS income limit. This means that the government site is wrong!

image

Because of that, I will no longer consider the maximum income limit when the client has Partial OAS. That reminds me... I need to consider when the _partner_ has Partial OAS.

- [x] consider when the partner has Partial OAS
- [x] more tests


utils/api/benefits/gisBenefit.ts
```diff
@@ -37,11 +38,19 @@ export class GisBenefit extends BaseBenefit {
...
-    const meetsReqIncome = this.income < maxIncome
+    const meetsReqIncome =
+      this.income < maxIncome ||
+      /*
+       This exception is pretty weird, but necessary to work around the fact that a client can be entitled to GIS
+       while being above the GIS income limit. This scenario can happen when the client gets Partial OAS, as
+       GIS "top-up" will come into effect. Later, in RequestHandler.translateResults(), we will correct for
+       this if the client is indeed above the true (undocumented) max income.
+      */
+      this.oasResult.entitlement.type === EntitlementResultType.PARTIAL
```
KhalidAdan, line 53: Sounds good 👍🏾 

`...above the true (undocumented) max income.` 😅

utils/api/benefits/gisBenefit.ts
```diff
@@ -112,11 +121,13 @@ export class GisBenefit extends BaseBenefit {
...
-      result === -1
-        ? EntitlementResultType.UNAVAILABLE
-        : EntitlementResultType.FULL
+    const result = roundToTwo(this.getEntitlementAmount())
+
+    let type: EntitlementResultType
+    if (result === -1) type = EntitlementResultType.UNAVAILABLE
+    else if (result === 0) type = EntitlementResultType.NONE
+    else type = EntitlementResultType.FULL
+
```
KhalidAdan, line 130: Much easier to read!

KhalidAdan (approved):

PR #197, opened 2022-02-28
Added normal whitespacing and a min width for the apply buttons

KhalidAdan:
Fixes the buttons on mobile


JeremyKennedy (approved):
Better on desktop, much better on mobile!

PR #202, opened 2022-03-01
Feature/feb28 design updates

KhalidAdan:
All of the quick wins brought up in yesterday's meeting. I'll make a separate PR for FAQs and one for each of the doozies (blur inputs on mobile and the cold start issue)


JeremyKennedy (approved):
Looks great! I Google Translated "Get Results" as I think we should try to stop using `FRENCH: ` placeholders. I've asked Wiam in Slack to confirm this wording.

PR #207, opened 2022-03-02
Change Enter to Done on mobile

JeremyKennedy:
On mobile, during user testing we saw users press the Enter button after they enter an age or an income. Previously, this would result in submitting the form, which would either fill the page with "this field is required" errors, or abruptly bring them to the results page. Now, the user's keyboard will be updated with a "done" button (a checkmark on Android), and pressing it will hide the keyboard rather than submitting the form.

The below screenshots show the difference in the keyboard's Enter button:

Before:

After:


KhalidAdan (approved):
Squash and 🚢

PR #209, opened 2022-03-02
Fix page reloading bug by switching to local implementation

JeremyKennedy:
By making 100% of the app logic local, this simplifies many things:
- infrastructure: now there is only static content to deploy, no need to worry about an API or serverless functions or cold starts
- frontend behavior: instead of having to make a network request and wait for a response, the frontend is now able to instantly process data, meaning the app processing is super fast, and needs no internet connection (once loaded)
- deployment: there's no fancy "static site generation" or "server site properties" or whatever fancy web magic Next.JS offers, just a simple static site

Initially we went with the backend API approach for app logic as we were considering databases, authentication, and whatnot, but that is all out of scope. Adobe Analytics is implemented and is all we need.


client-state/models/Form.ts
```diff
@@ -160,31 +156,20 @@ export const Form = types
...
   }))
   .actions((self) => ({
     sendAPIRequest: flow(function* () {
-      // build query  string
-      const queryString = self.buildQueryStringWithFormData()
-
-      const apiData = yield fetch(`/${self.API_URL}?${queryString}`)
-      const data: ResponseSuccess | ResponseError = yield apiData.json()
+      const input = self.buildObjectWithFormData()
+      const data = mainProcessor(input)
```
KhalidAdan: Do you know if CURAM is completely dead or not? I wonder if this will cause integration issues down the line. I'd hate for a dev to pick this up and realize they have an application where they need  to re-implement the API.

That being said, I _hate_ this loading issue and I think if CURAM really is gone then this is actually the best solution as it simplifies the app tremendously.

KhalidAdan (approved):
I'm okay with this, but just left a thought that we could chat about at any time really. The solution came out clean!

client-state/models/Form.ts
JeremyKennedy: I don't know definitively if Curam is forever out of the picture or not. I do know that if it is needed, it is still possible to use the existing API and build off that, but the more likely solution would be a new endpoint anyways, which means this change shouldn't affect Curam at all. The only case where this would mess things up is if we needed to communicate with some external system (db or Curam) on every frontend request. I don't that will be the case anymore, and a Curam button would be a new endpoint that's triggered by a new button in the UI.

client-state/models/Form.ts
```diff
@@ -165,6 +167,17 @@ export const Form = types
...
+    buildQueryStringWithFormData(): string {
+      const parent = getParent(self) as Instance<typeof RootStore>
+      let qs = `_language=${parent.lang}`
+      for (const field of self.fields) {
+        if (!field.value) continue
+        if (qs !== '') qs += '&'
+        qs += `${field.key}=${fixedEncodeURIComponent(field.sanitizeInput())}`
+      }
+      return qs
+    },
```
KhalidAdan, line 181: 👍🏾

PR #217, opened 2022-03-09
Fix tooltip height on mobile, block scrolling on mobile

KhalidAdan:


JeremyKennedy (approved):
Appears to work fine. On mobile, when a tooltip is open, scrolling will no longer scroll the page and only scrolls the tooltip.

PR #218, opened 2022-03-10
remove trailing breadcrumb

KhalidAdan:


JeremyKennedy (approved):
Looks good. I see you've added support for breadcrumbs to have links other than `/`, which should be useful once we add breadcrumbs for Questions/Results/etc!

PR #219, opened 2022-03-10
Removed alerts from the app

KhalidAdan:
First step of the refactor of the results and questions pages


JeremyKennedy:
I believe this is safe to close, as #222 does a better job of this (it does not just remove the alert and content, but removes the alert and presents the content differently).

PR #225, opened 2022-03-15
Update Tailwind to 3

JeremyKennedy:
This is the only major version update we are behind on. Unfortunately it does break some things:
- [x] tooltip icon is invisible
- [x] in tooltip, tooltip icon in header is inverted

### DIFF
--- a/components/Tooltip/tooltip.tsx
+++ b/components/Tooltip/tooltip.tsx
@@ -55,7 +55,7 @@ export const Tooltip: React.FC<{
         xmlns="http://www.w3.org/2000/svg"
         width="25"
         height="25"
-        fill="none"
+        fill="primary"
         className="cursor-pointer z-20"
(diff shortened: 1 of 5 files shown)


JeremyKennedy:
We have more important things to do than fixing a dependency update, but this would be really nice to get done, as it is the only major version we are behind.

KhalidAdan:
This one gives me the hereby jeebies

alex-solo (approved):
Met with Khalid who approves these changes

JeremyKennedy:
Hey Alex - looks great! One thing I noticed is that the tooltip icon is now black: 
image
it should be more of a blue: 
image

Inspecting element was a bit weird, it seemed like changing the color via the browser didn't actually affect the tooltip at all. Not sure exactly what is causing this.

JeremyKennedy:
Looks good, thanks!

PR #231, opened 2022-03-22
Question changes

JeremyKennedy:
Two small changes:
- move partner income question down, to match client ordering
- remove 'legal status other' question (this touched a bunch of files)


KhalidAdan (approved):
🚢

PR #238, opened 2022-03-25
added localized numbers

KhalidAdan:
Localizes income field


JeremyKennedy (changes requested):
Visually it looks good, but I have a feeling the sanitize function is not removing everything that it needs to. Also note when you do update that, it shouldn't just blindly remove commas, as in French, commas are the decimal. 
image

JeremyKennedy (approved):

PR #239, opened 2022-03-25
Added new table column

KhalidAdan:
Added new table column, if we want to change the API we can stop splitting the data coming from the benefit details


components/ResultsTable/index.tsx
```diff
@@ -99,6 +152,7 @@ export const ResultsTable = observer(() => {
           {!root.summary.zeroEntitlements && (
             <tr className="border-t border-content font-bold ">
               <td colSpan={2}>{tsln.resultsPage.tableTotalAmount}</td>
+              <td></td>
```
JeremyKennedy: image
if this line was removed, would it move the table footer into a single line? I think ideally that is how it would look

JeremyKennedy (commented):
I see no changes on mobile, however I think that is perfectly fine:
image

JeremyKennedy (approved):

PR #240, opened 2022-03-25
Added FAQ to eligibility page

alex-solo:
### Description

List of proposed changes:

-
-

### DIFF
--- a/components/FAQ/index-e.tsx
+++ b/components/FAQ/index-e.tsx
@@ -2,545 +2,571 @@ import Link from 'next/link'
 
 export const FAQE = () => (
   <div>
-    <h2 className="h3 mt-10">Old Age Security</h2>
     <details>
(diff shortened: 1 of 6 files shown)


KhalidAdan:
Looks pretty good! Tailwind in two days is pretty fast Alex! Last day is tomorrow so I'll leave it to Jeremy to give the thumbs up!

JeremyKennedy (changes requested):
Yeah, looks good! I think the only thing missing is to add this exact same component to the bottom of the results page. Make sure it is visible even when the estimation results is unavailable (ie. when legal status is other).

JeremyKennedy (approved):
Awesome, thanks! Please note the commit I just added, any text strings need to be pulled from the translation files. Small change but important thing to remember going forward!

PR #242, opened 2022-04-04
fix for fitting text inside button on results page

alex-solo:
### Description

List of proposed changes:

-
-


JeremyKennedy (approved):
Simple and works great!

PR #251, opened 2022-04-13
Updated design system to latest version and added several components (Button, Breadcrumb, ErrorPage)

alex-solo:
### Description

List of proposed changes:

-
-

### Additional Notes
adding multiple branches together


JeremyKennedy (approved):
Looks good! I made a few changes, pretty much just small cleanup stuff (one was removing `package-lock.json`.

alex-solo:
@JeremyKennedy The 404 page looks not great with the breadcrumbs and h1 still there which I removed. We can talk about modifying the Layout component maybe or just having a bare 404 page like this: https://www.canada.ca/errors/404.html

Also using the Layout as is, the "main" section needs to lose the "mainContent" class to get the line-height correct and fix the underline issue.

alex-solo:
Error page also needs a different footer ("Error Page Footer")

PR #252, opened 2022-04-14
cleanup and fixes

JeremyKennedy:
This PR is mainly a bunch of small unrelated stuff,  because of this I recommend reviewing this by commit, rather than using GitHub's "files changed" view.

Changes:
- clean up faq page layout, fix faq page link in need help
- fix issue saying eligible when not due to high income
- .idea: add tailwind's className to allowed html attributes
- replace detailOverride method with more straightforward one
- split results table row into components
- fix invalid nesting of `<p>` elements
- organize imports


components/Forms/ComponentFactory.tsx
```diff
@@ -193,10 +193,7 @@ export const ComponentFactory: React.VFC = observer(({}) => {
         <NeedHelp title={tsln.needHelp} links={root.summary.needHelpLinks} />
       </div>
 
-      <div className="mt-16">
-        <h2 className="h2 text-content mb-8">{tsln.faq}</h2>
-        <FAQ />
-      </div>
+      <FAQ />
```
MarcoGoC, line 196: for accessibility this should be an abbr

MarcoGoC (approved):
too many changes :) approved to move along but hard to validate everything. 
see comment where an abbr should be added

components/Forms/ComponentFactory.tsx
JeremyKennedy: (will follow up on this, just want to merge the PR now as it affects a bunch of files)

JeremyKennedy: what do you mean should be an abbr? I'm not familiar with that term in this context.

MarcoGoC: nevermind I thought it was text but it is a component

PR #254, opened 2022-04-14
Added footer from the component library

alex-solo:
### Description

The "upArrow" image will not appear until the following PR in the Design System is merged: https://github.com/DTS-STN/DECD-Design-System/pull/235

List of proposed changes:

-
-

### DIFF
--- a/components/Layout/Footer.module.css
+++ b/components/Layout/Footer.module.css
@@ -1,3 +1,0 @@
-.footerBackground {
-  background: #26374a url(/bgc.png) no-repeat 100% 100%;
-}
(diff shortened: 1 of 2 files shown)


alex-solo:
The design system PR that this PR depends on has been approved and merged.

MarcoGoC (dismissed):
we should add the links, and probably have them on the translation file 
btnLink should point to the H1 of the page

JeremyKennedy:
Definitely not a bad idea to add actual links. It could be a separate PR at a later time, up to you. But for an idea of how we've previously handled links that need to be translated, check out the backend translation file. We have a bunch of links defined there.

JeremyKennedy (approved):
Definitely not a bad idea to add actual links. It could be a separate PR at a later time, up to you. But for an idea of how we've previously handled links that need to be translated, check out the backend translation file. We have a bunch of links defined there.

PR #249, opened 2022-04-11
Added Breadcrumb and ErrorPage components

alex-solo:
### Description

List of proposed changes:

-
-

PR #258, opened 2022-04-19
Added Link component

alex-solo:
### Description

List of proposed changes:

-
-

### DIFF
--- a/components/NeedHelp/index.tsx
+++ b/components/NeedHelp/index.tsx
@@ -1,4 +1,4 @@
-import { Link } from '@dts-stn/decd-design-system'
+import { Link as DSLink } from '@dts-stn/decd-design-system'
 import { Instance } from 'mobx-state-tree'
 import { useEffect } from 'react'
 import { SummaryLink } from '../../client-state/store'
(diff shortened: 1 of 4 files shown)


MarcoGoC (approved):
straight forward

alexunique0519 (approved):
LGTM

JeremyKennedy (changes requested):
Looks pretty good! The main thing I noticed is that links need to open in a new tab. This is what `target="_blank"` is doing. That used to be in Need Help but was removed, however it looks like it was never in More Info (which it should have been before).

Anyway, please see if adding `target="_blank"` to all of these links works, we want them to open in a new tab so that users don't lose their data entered into the app.

Also, for consistency and clarity, since Next has its own `Link` component, could we make all design system links imported as `DSLink`?

alex-solo:
@JeremyKennedy Thanks Jeremy. I made the suggested changes and also made changes to the Link component in the design system. That PR can be found here: https://github.com/DTS-STN/DECD-Design-System/pull/249

JeremyKennedy (approved):
Looks good! I rebased onto develop and then updated the design system since your PR was merged. Now, these links open in a new tab!

I noticed there are some other links in the application that weren't migrated to the `DSLink` component, but that's not the end of the world... I found them by searching the codebase for `text-default-text underline`. You can change them if you like, it's mostly just stuff within the FAQ.

Also, when merging, please use a "squash" commit as in this case we don't need to preserve the commit history of the PR.

PR #259, opened 2022-04-21
 adding top menu/translations

MarcoGoC:
### Description

List of proposed changes:
- Added 'hidden' top navigation / enabled when using the keyboard to navigate the page

- 

### Additional Notes
- we are probably going to remove the switch to basic html link, but it is there for now.


JeremyKennedy (changes requested):
I can see the added component, very cool! Couple things:
- I feel like we can get rid of the "about gov" and "basic html" pieces before merging this PR
- from what I could tell, the skipping to main content doesn't work. Perhaps the href should be updated to `#applicationTitle` which is the id for the h1/heading component on the develop branch? (you would need to rebase this branch onto develop to see that new id)

MarcoGoC:
Agree on the html, I believe the about government is a mandatory element.

i18n/web/index.ts
```diff
@@ -25,6 +25,9 @@ export type WebTranslations = {
   langLong: string
   creator: string
   search: string
+  skipToMain: string
+  skipToAbout: string
+  switchToBasic: string
```
JeremyKennedy, line 30: the basic html part can be removed here, now that we aren't using it

components/Layout/index.tsx
```diff
@@ -22,6 +22,21 @@ export const Layout: React.VFC<{
...
+        <ul id="applicationTitle" className="z-10">
+          <li className="left-0 top-2 absolute text-center box-border w-full -z-50 focus-within:z-50 ">
+            <a className="focus:bg-[#26374a] font-bold p-1 text-white" href="#wb-cont">{tsln.skipToMain}</a>
+          </li>
+          <li className="left-0 top-2 absolute text-center box-border w-full -z-50 focus-within:z-50 ">
+            <a className="focus:bg-[#26374a] font-bold p-1 text-white" href="#wb-info">{tsln.skipToAbout}</a>
+          </li>
+          {/* <li className="left-0 top-2 absolute text-center box-border w-full -z-50 focus-within:z-50 ">
+            <a className="focus:bg-[#26374a] font-bold p-1 text-white" href="?wbdisable=true" rel="alternate">{tsln.switchToBasic}</a>
+          </li> */}
```
JeremyKennedy, line 36: this commented code can be removed

components/Layout/index.tsx
```diff
@@ -22,6 +22,21 @@ export const Layout: React.VFC<{
     <>
       <HeadDoc />
       <SCLabsTestHeader />
+
+      <nav>
+        <ul id="applicationTitle" className="z-10">
+          <li className="left-0 top-2 absolute text-center box-border w-full -z-50 focus-within:z-50 ">
+            <a className="focus:bg-[#26374a] font-bold p-1 text-white" href="#wb-cont">{tsln.skipToMain}</a>
```
JeremyKennedy, line 29: the href here should be changed to `applicationTitle` so the link works

components/Layout/index.tsx
```diff
@@ -22,6 +22,21 @@ export const Layout: React.VFC<{
     <>
       <HeadDoc />
       <SCLabsTestHeader />
+
+      <nav>
+        <ul id="applicationTitle" className="z-10">
```
JeremyKennedy, line 27: this id should be removed or changed, as it will conflict with the `applicationTitle` element [here](https://github.com/DTS-STN/eligibility-estimator/blob/develop/components/Layout/index.tsx#L56)

components/Layout/Footer.tsx
```diff
@@ -7,7 +7,7 @@ import styles from './Footer.module.css'
 export const Footer = () => {
   const tsln = useTranslation<WebTranslations>()
   return (
-    <footer className="text-[14px]">
+    <footer id="wb-info" className="text-[14px]">
```
JeremyKennedy, line 10: this id could probably be changed to something more relevant, like `footer` (note to change the href in the layout index file too)

JeremyKennedy (changes requested):
I added some comments that I think should get this working, as the skip to main content link doesn't work currently

PR #262, opened 2022-05-02
formulas for entitlement

JeremyKennedy:
This PR changes the GIS/ALW/AFS entitlement calculations to used formulas, rather than using the tables posted on Canada.ca.

An important thing to note while reviewing is that this logic is heavily tested. The fact that no tests changed/failed should help confirm that this logic works exactly as the previous logic did.


alex-solo (approved):
I will trust the tests!

PR #263, opened 2022-05-10
AccordionForm implementation

alex-solo:
…tate## [DC-XXX](https://jira-dev.bdm-dev.dts-stn.com/browse/DC-XXX) (Jira Issue)

### Description

List of proposed changes:

-
-


alex-solo:
2 design system PRs that will make it work. For now, I am using a copy of the design system with the necessary fixes. Once the PRs are approved, we can remove this package and update the official one. For now, we can at least see the fixes in action.

https://github.com/DTS-STN/DECD-Design-System/pull/294
https://github.com/DTS-STN/DECD-Design-System/pull/295

alex-solo:
Another Design System PR that needs to be approved, merged and deployed in the next version. This one addresses the flickering button on the last card: https://github.com/DTS-STN/DECD-Design-System/pull/298

components/Hooks/index.tsx
```diff
@@ -1,5 +1,5 @@
 import { useRouter } from 'next/router'
-import { useCallback, useContext, useEffect, useState } from 'react'
+import { useCallback, useContext, useEffect, useState, useRef } from 'react'
```
JeremyKennedy: unnecessary added import?

alex-solo:
Once a new package is released that has this PR, the flickering button issue will go away: https://github.com/DTS-STN/DECD-Design-System/pull/298

JeremyKennedy (approved):
Amazing! Very happy to see this work without the major mobx removal.

There was only one issue I had to solve - the card was assumed valid if ANY field was filled, rather than all fields. This meant that you could fill `age` and skip `oasAge`, and it would let you proceed. I fixed that by tweaking the validation logic, so now it runs validation on all fields on change.

Unfortunately the above change requires DS PR approval: https://github.com/DTS-STN/DECD-Design-System/pull/301

Perhaps we should switch to a `solosphere` release while we wait for their approval and release?

I also took this opportunity to improve/refactor/clean things, such as:
- adding more types (this really helps with readability and understanding the code)
- other readability improvements, like reducing nested functions, adding comments, etc.
- moving the category/step definitions to the backend, where we already define categories. I also refactored field ordering to ensure the order is coming from the backend.
- renamed `canadaWholeLife` to `livedOutsideCanada`, to better match the updated question (this affected many lines!), plus switched back to a bool rather than a yes/no enum
- cleaned up the tooltip definitions, now there's no `showTooltip` parameter, and it'll just show it when it's defined
- removed some other unnecessary/redundant code

PR #267, opened 2022-05-11
 SAEB 1266 - Starting again Header

MarcoGoC:
### Description
Using the Header from the design center

List of proposed changes:
- Updated Id for the footer
- Removed breadcrumbs-variable from 404
- Updated Layout to use the header
- Removed unused Header.jsx

- download and try


alexunique0519 (approved):
LGTM

PR #268, opened 2022-05-13
Saeb 1267 update intro

alexunique0519:
## [DC-XXX](https://jira-dev.bdm-dev.dts-stn.com/browse/DC-XXX) (Jira Issue)
https://jira-dev.bdm-dev.dts-stn.com/browse/SAEB-1267
### Description
Make changes to the intro page to follow the final mockups, also replace the footer with the footer component from DECD-Design-System 

All missing links have been added.

List of proposed changes:

-
-

### Additional Notes
The breadcrumb was missing due to the Header component (from design system) 's prop name change. Added it back with using a latest prop name.


components/Layout/Footer.tsx
```diff
@@ -4,72 +4,6 @@ import { useTranslation } from '../Hooks'
...
-  ]
 
-  return (
-    <DSFooter
-      btnLink="#"
-      landscapeLinks={landscapeLinks}
-      brandLinks={brandLinks}
-    />
-  )
+  return <DSFooter btnLink="#" id="footer" lang={tsln._language} />
```
MarcoGoC: need merge with develop to add the id

MarcoGoC (approved):

i18n/web/en.ts
```diff
@@ -1,12 +1,14 @@
 import { WebTranslations } from '.'
 import { Language, Locale } from '../../utils/api/definitions/enums'
 
+import api_translation from '../api/en'
```
JeremyKennedy: is this import used? It doesn't seem to be. if not, please remove. if it is used, please use `camelCase` for all variable names!

i18n/web/fr.ts
```diff
@@ -1,13 +1,14 @@
 // noinspection SpellCheckingInspection
 import { WebTranslations } from '.'
 import { Language, Locale } from '../../utils/api/definitions/enums'
+import api_translation from '../api/fr'
```
JeremyKennedy: same comment as previous

JeremyKennedy (changes requested):
small change to make with importing the api translations into the web translations

i18n/web/en.ts
alexunique0519: Yes, it is used for those link URLs. I have changed the variable names

i18n/web/fr.ts
alexunique0519: Yes, it is used for those link URLs. I have changed the variable names

MarcoGoC (approved):

alex-solo:
breadcrumbs missing, waiting on update

tailwind.config.js
```diff
@@ -18,14 +18,16 @@ module.exports = {
       h5: ['20px', '22px'],
       h4: ['22px', '24px'],
       h3: ['24px', '26px'],
-      h2: ['32px', '36px'],
-      h1: ['34px', '48px'], // requires red <hr /> below
+      h2: ['36px', '40px'],
+      xs: ['32px', '36px'],
+      h1: ['38px', '42px'], // requires red <hr /> below
```
JeremyKennedy, line 23: These changes are global, and affect the whole application - I am not sure if this is intended (or good). If it is intended, then I suppose that's fine. If not, could we see if we can apply these styles within the component, rather than globally?

styles/globals.css
```diff
@@ -149,10 +154,33 @@
...
+    @apply sm:w-8/12
+  }
+
+  #mainHeader > header > div {
+    padding-left: 0px !important;
+    padding-right: 0px !important;
+  }
+
+  #mainHeader > header > div.ds-container > nav > ul > li {
+    margin-left : 0px !important;
```
JeremyKennedy: It looks like these CSS changes apply to the Design System header. We should not be styling DS components, this is fragile and not the best way of doing things. If the DS component needs styling, we should open a PR with them, so that the fix applies to all users of the DS, and there is no risk of the style breaking as things are updated.

JeremyKennedy (changes requested):
I made some very small commits to clean some things up, and added some questions on CSS. It is looking really good though!

tailwind.config.js
alexunique0519: I think we need those changes. The change for h1/h2 font is to make sure styles align with final mockups, where h1 is defined as font size 38px, line-height 42, and h2 has 36px font size and 40px line height

styles/globals.css
alexunique0519: Sure, I will revert this change and create a PR on design-system repo

JeremyKennedy (approved):

PR #274, opened 2022-05-17
fix bug with marital statuses, add test to catch

JeremyKennedy:
This fixes a bug when switching between marital statuses. 

To reproduce, fill the form with age 65, canadian citizen, canada whole life, income 10, single. Then change marital to married. Then change it to separated. It should hide the questions when separated but it does not.


MarcoGoC (approved):
+1

PR #275, opened 2022-05-17
Add "involuntarily separated" and merge marital statuses

JeremyKennedy:
This PR merges the marital status into as few options as possible. It also adds "involuntarily separated" as an option, which we will currently not provide an estimation for. Selecting it will display an error.


MarcoGoC (approved):

alex-solo (approved):
simplifies things, nice!

PR #277, opened 2022-05-18
Saeb 1231 wcag

alexunique0519:
### Description

List of proposed changes:
- Added title for SVGs 
- Added missing aria-labels
- Always focus on the first error on the page when users click "Get Results"
- Replaced the old footer with the footer from the design-system lib
- Fixed the issue that focus is missing on radio options. 
- Fixed the issue that an "Enter" button click could not select a radio option
-


JeremyKennedy:
Converted to draft as I just merged in the accordion PR, which caused conflicts here and very likely broke things. Once resolved, please mark this as ready for review!

If some things are hard to re-implement, like scrolling to errors, no worries, just take those out and we can merge this in without that.

MarcoGoC (approved):

PR #282, opened 2022-06-06
Abort early when legal status is not supported

JeremyKennedy:
As per the mockups, we should be displaying an error to the user as soon as we know we won't be able to provide an estimation. This PR does this for Legal Status when it is Sponsored or Other. To do in another PR will be the more difficult logic when the client hasn't lived in Canada for long enough.


MarcoGoC (approved):

alex-solo (commented):
works as expected!

PR #283, opened 2022-06-06
Remove table-based entitlement calculations

JeremyKennedy:
We have moved to using formula-based calculations. This was kept in case there were issues, but now it's causing problems, so let's remove it.

Note that this has zero impact on how the app works.


MarcoGoC (approved):
some changes seem the same as the next PR

alex-solo (approved):

PR #285, opened 2022-06-06
Dependency updates and cleanup

JeremyKennedy:
This update does a number of cleanup tasks:
- Updating some core dependencies (Yarn, Node, ESLint)
- Pin `next` to `12.1.4`, since future versions break things (the unreleased `12.1.7` will be okay)
- Update Jest config to recommended from Next.js
- Get rid of `swagger-ui-react` as a dependency, and instead serve that page as static html, pulling the dependency from a CDN
- Remove several unused dependencies
- Add several missing dependencies used for type checking


MarcoGoC (approved):

alex-solo (commented):
No more swagger UI means no more being able to play around with the API? If it's too far out of sync, that makes sense and it should get rid of that dependabot alert?

JeremyKennedy:
> No more swagger UI means no more being able to play around with the API?

There is still an API page, the change here just makes it so we are using a CDN plus static HTML to host it, rather than bundling it into our application directly. This reduces our dependencies by a lot!

> If it's too far out of sync, 

It is a little bit out of sync now, but I do have a WIP PR to bring this stuff back up to date. #272 

> and it should get rid of that dependabot alert?

It sure will!

PR #286, opened 2022-06-06
Modern Yarn: Plug and Play + Zero-Install

JeremyKennedy:
This PR updates our project to use two modern Yarn features, Plug and Play (PnP), and Zero-Install.

PnP essentially moves from using a `node_modules` folder to using archives. This makes install performance a lot faster, since there aren't hundreds of thousands of files to copy. PnP can sometimes cause issues with IDEs and tools, though it should be mature enough by now, and is even the default in Yarn nowadays. Please let me know if you have any issues, like with ESLint, Jest, or Prettier in your IDE. More info: https://next.yarnpkg.com/features/pnp

Zero-Install essentially stores the archives above in our repo directly. This means that you never have to run `yarn install`, even when switching branches! It also means that our deployment infrastructure doesn't need to pull any dependencies, since they're all stored in the repo, making deployment faster and more reliable. More info: https://yarnpkg.com/features/zero-installs


MarcoGoC (approved):

MarcoGoC (approved):
this is the only project that uses yarn

alex-solo (approved):
great update

PR #293, opened 2022-06-13
fix: updated latest design system lib and make change to footer and h…

alexunique0519:
### Description

The header and footer components have been changed to remove the padding. The issue has been resolved in this PR. 

List of proposed changes:
- Update the design system library to the latest version.
- Pass new props to header and footer component to remove the padding


JeremyKennedy (approved):
looks good!

PR #299, opened 2022-06-14
Clean up linting and workflows

JeremyKennedy:
This PR is just a few smaller changes:
- Integrate Prettier into ESLint's checks. This makes unformatted code fail our automated tests!
- Add more files to `.prettierignore` (stuff that should have never been processed by Prettier)
- Make CodeQL analysis run on `develop`. This was causing problems when only running on `main`, because we would merge things into `develop` and only realize they were issues when doing a prod release. Now, if there are issues, we'll see right away.
- Update the versions of the actions used in our workflows


MarcoGoC (approved):
straight forward

alex-solo (commented):
nice changes!

PR #304, opened 2022-06-17
Fix: fixed the screen size order in tailwind config

alexunique0519:
### Description

I found our media query CSS rules not working properly in some cases, then I double-checked the tailwind config, and realized that the screen size order was wrong. 

List of proposed changes:

- reorder the screen size from xs to lg


JeremyKennedy (approved):
looks good, thank you for finding and fixing this!

PR #305, opened 2022-06-20
Legal Values cleanup

JeremyKennedy:
This PR removes the legacy "intermediary" object for accessing Legal Values, and simply accesses them directly from the JSON file. I've also cleaned up some of the naming conventions in the JSON file.


alex-solo (approved):
looks good

utils/api/benefits/entitlementFormula.ts
```diff
@@ -5,7 +5,7 @@ import {
   PartnerBenefitStatusHelper,
 } from '../helpers/fieldClasses'
 import roundToTwo from '../helpers/roundToTwo'
-import { legalValues } from '../scrapers/output'
+import legalValues from '../scrapers/output'
```
MarcoGoC, line 8: this line show as error on VS code.  
"... did you mean to use { } ... instead"

MarcoGoC (commented):
weird VS code shows all these lines as error, maybe I am missing something  
  import legalValues from '../scrapers/output'

utils/api/benefits/entitlementFormula.ts
JeremyKennedy: I think the VSCode errors is just a caching issue. If you load `/scrrapers/output/index.ts` it should update the VSCode error checking cache, I think!

PR #306, opened 2022-06-20
support when income is not known

JeremyKennedy:
This PR mainly adds support for when income is not known. This is done by first asking the user if they want to provide their income, and if they do, then we show the income question. Same for partner income. If either income is not provided, that will affect the results available.

I have differed from the mockups somewhat in order to accommodate the existing application structure. Rather than having a checkbox and an optional field, I've gone with radio buttons which will show the income field if Yes is selected (exactly as we have the OAS Deferral). This keeps consistency with the rest of the app, and keeps complexity to a minimum, since we're using logic our app supports already.

The results page is still using the old layout, as the new one is not complete yet.

Several things were improved as a result of this PR:
- compressed tests some more, using variable unpacking
- update text replacement rules to optionally have the context of their benefit, allowing us to know what benefit we're working with in the replacement logic (this simplified some existing rules, and allowed me to include the max income in the detail when income is not known)
- updated the detail wording for a few things, like saying "you and your partner's combined income" instead of "your income" when partnered
- force-capitalize each line of the eligibility detail text, required due to the above change
- fix field sorting not actually working in the ui
- reduce debounce for currency fields, to increase responsiveness. the reason we had it was to prevent pop-in and errors as the user types, but with our new layout, neither of those things happen (for currency fields - not the case for age!)

If interested, you can check out some screenshots I've added to JIRA here: https://jira-dev.bdm-dev.dts-stn.com/browse/SAEB-1310


alex-solo (approved):
Demoed, everything works!

utils/api/definitions/enums.ts
```diff
@@ -1,3 +1,10 @@
+export enum BenefitKey {
+  oas = 'oas',
+  gis = 'gis',
+  alw = 'alw',
```
MarcoGoC, line 4: alw or allowance ?

MarcoGoC (approved):

utils/api/definitions/enums.ts
JeremyKennedy: alw is the short form // abbreviation we're using for allowance through the app! it's actually used by the gov in some internal documents I saw as well (which is actually what prompted me to use alw in the first place)

PR #308, opened 2022-06-24
fix gis calc when partial oas and high income

JeremyKennedy:
When Partial OAS (years in canada = 20) and high income (100k), GIS would be negative. Now we will default to zero in that case (though the final result is not zero).


MarcoGoC (approved):
I am not sure what the result should be, I get  a negative amount ? 

Old Age Security (OAS) | $324.33
Guaranteed Income Supplement (GIS) | -$3,022.87
Total | -$2,698.54

PR #314, opened 2022-06-27
 i18n rewrites

MarcoGoC:
### Description

List of proposed changes:
- Added rewrites to config.next and links 

- visit the pages and check the links
- answers links back to the question **are not** working for some cases


pages/index.tsx
```diff
@@ -1,4 +1,5 @@
 import { Button } from '@dts-stn/decd-design-system'
+import Link from 'next/link'
```
alexunique0519: Looks like this Link component has been imported but not being used anywhere

components/ResultsPage/YourAnswers.tsx
```diff
@@ -37,7 +37,7 @@ export const YourAnswers: React.VFC<{
               <strong>{getDisplayValue(input)}</strong> &nbsp;
               <DSLink
                 id={`edit-${fieldKey}`}
-                href={`/eligibility#${fieldKey}`}
+                href={`${tsln.eligibility}#${fieldKey}`}
```
alexunique0519: On French mode, the link could not navigate to the accordion form, instead, it shows "we could not find the web page"

MarcoGoC: ya some of the links from the answers dont work as the id's change but it is a different issue

PR #315, opened 2022-06-28
Tooltip and mobile layout change

alexunique0519:
### Description
The tooltip icon is replaced with a more information link. The layout dimensions have been changed for mobile view.

List of proposed changes:
- Tooltips icon is replaced with a link
- Extended the main content width for mobile views
- Fixed the "what you told us" column issue


components/Forms/Radio.tsx
```diff
@@ -33,7 +33,7 @@ export const Radio: React.VFC<InputProps> = observer((props) => {
 
   return (
     <div className="radio">
-      <div className="mb-2.5">
+      <div className="">
```
habibaz: do we need to keep this div

components/Forms/Radio.tsx
```diff
@@ -45,17 +45,21 @@ export const Radio: React.VFC<InputProps> = observer((props) => {
             dangerouslySetInnerHTML={{ __html: label }}
           ></span>
         </label>
-        <span>
+        {/* <span>
```
habibaz: if this is not needed can we please remove this

alexunique0519: Yes, I just reverted this change, I will create a separate PR to address that change

components/Forms/Radio.tsx
alexunique0519: Yes, otherwise the "more information" link will come to the end of this line. I deleted the "className".

alex-solo:
Screen Shot 2022-06-28 at 3 17 55 PM

I will approve this PR since it wasn't addressing the issue in the image above, but it's something to fix here or in a separate PR.

alex-solo (approved):
Please see comment I left

JeremyKennedy (changes requested):
In general it would be better to split PRs into specific items, by grouping unrelated changes in, it's a bit harder to review/test. Now since I think we'll be going with a different tooltip/more-info design, the other changes in this PR will have to wait until the redesign is done. Though, we could pull out the unrelated changes and merge those in while the tooltip redesign is in progress... 🤔

JeremyKennedy:
I've split this PR into two separate ones, #317 and #318. Together, those should be exactly the same as this one, so I will close this one.

PR #316, opened 2022-06-28
Fix "screen out" behaviour

alex-solo:
### Description

Jira task SAEB-1311

List of proposed changes:

-
-

### DIFF
--- a/__tests__/pages/api/benefits.test.ts
+++ b/__tests__/pages/api/benefits.test.ts
@@ -7,6 +7,7 @@ import {
   ResultKey,
   ResultReason,
   SummaryState,
+  ValidationErrors,
 } from '../../../utils/api/definitions/enums'
 import { FieldKey } from '../../../utils/api/definitions/fields'
 import roundToTwo from '../../../utils/api/helpers/roundToTwo'
@@ -24,7 +25,6 @@ import {
   expectOasEligible,
   expectOasGisEligible,
   expectOasGisTooYoung,
-  expectOasGisUnavailable,
   income10k,
   partnerIncomeZero,
   partnerNoHelpNeeded,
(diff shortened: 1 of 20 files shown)


alex-solo:
@JeremyKennedy @MarcoGoC @alexunique0519 Any ideas what we want to do with these tests? Now that we "screen out" when we're under 10 years in Canada, we don't get "unavailable" but get the error and cannot proceed. Doesn't look like we're testing for that.

JeremyKennedy:
Converted to draft as there is still work to be done (tests). I'll try to check 'em out soon!

JeremyKennedy:
I think I'm all done with this now. I had to fix a few things along the way.
- fixed ENTITLEMENT_AMOUNT text replacement using the sum for all benefits rather than just the current one
- properly handled involuntarily separated for eligibility (not entitlement! that's hard)
- fix AGREEMENT_COUNTRIES enum containing undefined items

I also refactored the implementation a bit to better support the `everLivedSocialCountry` question. If they select true, we'll tell them to contact service Canada. If they select false, we'll say they're not eligible because they haven't lived in Canada long enough.

Note that ALW and AFS are not affected by the "involuntarily separated" change, because they are only for people who are partnered or widowed.

Note: the screening out for "years in canada" / "ever lived social country" is more complicated than I thought. It isn't simply "require 10 years if you live in canada, 20 years otherwise" - it depends on the benefit. Because of this, there is no golden rule we can apply to screen out if they are in the mysterious 10-20 year range (since it's per benefit). So, we'll let them proceed to the results page as long as they're over 10 years at which point we will say results are unavailable for certain benefits.

Some pictures and more details on changes available here: https://jira-dev.bdm-dev.dts-stn.com/browse/SAEB-1311

JeremyKennedy (approved):

PR #319, opened 2022-06-28
update social agreement country link

JeremyKennedy:
Looks like Canada.ca moved a link. I've updated it to a working one now.


MarcoGoC (approved):
+1

PR #320, opened 2022-06-30
 changes to settings.kts

MarcoGoC:
Updates to settings.kts


JeremyKennedy (approved):
Hoping this doesn't cause issues as our initial DTS infra setup used `coabe`, which is the "correct" application name. Our repo is still using the original name. You could just change the repo, but changing everything to be consistent with the original name is fine with me too.

PR #325, opened 2022-07-05
Remove mobx

JeremyKennedy:
**note: I've tested this decently well, but would love thorough testing from others as well**

This is a huge rewrite, though I am confident it will make future maintenance and debugging easier. Look at all the files that were deleted, and relatively how little was added! And since we don't need mobx anymore, we can easily use TypeScript types throughout the frontend, rather than those funky mobx types.

Now, only the inputs are saved to the session storage, rather than the entire form and results. This reduces complexity when changing languages. 

I did some small cleanup with Locale/Language/LanguageCode, so really all you need to care about is just the Language now, no need to pass anything else around (like for `numberToStringCurrency`).

I also renamed a few things, like FieldData to FieldConfig, but I did this as little as possible to simplify the diff/review for this PR.

I recommend reviewing this with "hide whitespace" on for clarity: 
image


MarcoGoC (approved):
it looks good to me and i haven't found any issues 
thanks

alexunique0519 (approved):

PR #326, opened 2022-07-05
New dockerfile and Helm files

MarcoGoC:
### Description

List of proposed changes:
- update the helmfile names and folder names 
- update to the settings.kts to match teamcity


JeremyKennedy (approved):
You have some really weird branching going on which made the diff in GitHub inaccurate, so I had to review locally to see the true changes. After doing that I can see it's just a simple rename, so fine with me!

PR #327, opened 2022-07-05
remove 'boolean' question type, fixing bug

JeremyKennedy:
alternatively we could support the boolean question type, though it may not be worth the complexity


MarcoGoC (approved):
thanks

PR #337, opened 2022-07-12
complete previous change: remove boolean question type

JeremyKennedy:
I forgot to remove these parts in my previous PR.

PR #339, opened 2022-07-13
Change from Age to Birth Month+Year

JeremyKennedy:
Reopening from #302. 

Switches from a single Age field for age (and partner age) to a month and year field. This uses the Design System components. I decided against using Month/Year for the OAS Deferral age, since if someone thinks "oh I want to receive OAS when I'm 70 to get the most money!" then they would have to do math to figure out at which year they turn 70. So I just stuck with age.

It looks like this:
image

The heights are a little funky, that's been reported to the design system team.

It's worth mentioning that on the Results page, what we'll show here is actually the age, not the month/year:
image
This feels helpful to me, so users can easily see what age was calculated for the results, and should make things make more sense without them having to do math.

Note, needs #337 and #325 merged first.


MarcoGoC (approved):
works ok, 
The only issue I see when coming back from results to the question, the previously entered year is 'lost'

JeremyKennedy:
Yep good catch, that's exactly the case unfortunately. It needs a change in the design system that would allow us to pass values to the component. Currently we can't do that as far as I can tell, so it is the only component that doesn't work when you go back or reload. I might end up completing this change in the design system code since it is pretty important for us and it might take some time before they can get to it. Though I would still like to merge this in just to get things settled, even though this is still an issue.

PR #340, opened 2022-07-13
Misc Improvements

JeremyKennedy:
A bunch of small changes I've been meaning to make, that either clean things up, or resolve warnings.

See commit message for more detail.

Note, needs #337, #325, #339 merged first.


MarcoGoC (approved):
Looks good to me, although not sure about the scripts being removed from _app and also from _document.

JeremyKennedy:
Yeah good catch, I can explain those better. So it turns out `_document` is server-side only, so any of the adobe scripts there would never have worked. Next.js warned about this on build but we've been ignoring it for ages. So, I've moved the scripts to the suggested location by Next.js, however there seems to be some other issue with them as they are causing errors on the page, so I've commented them out temporarily. So basically, it wasn't working before because it was in the wrong spot, and it's not working now because it's commented (but it's at least in the right spot).

PR #341, opened 2022-07-13
result page wording changes when income is not entered

alexunique0519:
### Description
Wording changes to provide better support when income is not entered. 

List of proposed changes:
- Message box heading: change to: You may be eligible for benefits
- Message box body: change to: Depending on your income, you may be eligible for old age benefits. See the details below for more information.
- First Heading: change to: You may be eligible at this time: okay as-is
- Body for the above heading: change from "Based on your information, you may be eligible for:" to "Depending on your income and based on your information, you may be eligible for:"
- Second Heading: Your estimated monthly total is: okay as-is
- Body for the above heading: change from "Based on the information you've provided, you should expect to receive around $695.37 per month." to "Based on the information you've provided, you should expect to receive around $695.37 per month. However, this amount may be lower or higher depending on your income."
- OAS box : "You are likely eligible for this benefit if your income is less than $133,527. Depending on your income, you should expect to receive around $1,664.23 every month."

### DIFF
--- a/components/ResultsPage/EstimatedTotal.tsx
+++ b/components/ResultsPage/EstimatedTotal.tsx
@@ -2,7 +2,7 @@ import Image from 'next/image'
 import { useRouter } from 'next/router'
 import { getTranslations, numberToStringCurrency } from '../../i18n/api'
 import { WebTranslations } from '../../i18n/web'
-import { Language } from '../../utils/api/definitions/enums'
+import { Language, SummaryState } from '../../utils/api/definitions/enums'
 import { BenefitResult, SummaryObject } from '../../utils/api/definitions/types'
 import { useTranslation } from '../Hooks'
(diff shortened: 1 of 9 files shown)


JeremyKennedy (changes requested):
Need to differentiate between "available ineligible" and when income is not provided. In this PR you've changed it so if someone is ineligible, it will say they may be eligible depending on income. I probably need to provide more detail in this comment, though I may be able to tweak this before tomorrow...

JeremyKennedy:
Note that PR https://github.com/DTS-STN/eligibility-estimator/pull/316 will mean that this PR needs to move from `ENTITLEMENT_AMOUNT` to `ENTITLEMENT_AMOUNT_FOR_BENEFIT`.

JeremyKennedy (approved):
I've made changes so that the new text only shows up when the income is not known. Before my changes, this applied even when ineligible.

PR #343, opened 2022-07-15
Installed new package name from the DS

MarcoGoC:
### Description

List of proposed changes:
- Updated the import new name in all pages 
- replaced the old name with the new one in package.json and with the new version 1.48

- download branch, run yarn install and run


alexunique0519 (approved):
LGTM

PR #346, opened 2022-07-15
fix: result page refresh summary issue

alexunique0519:
### Description
Fixed the issue that refreshing the result page and getting a different summary. 

List of proposed changes:
- Disable ssr on ResultsPage component. 

Enter all data in the accordion form and get the estimated result on the result page first. Then, try to refresh the page to check if the summary changes.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,15 +1,29 @@
 {
+  "version": "0.2.0",
   "configurations": [
     {
-      "type": "node",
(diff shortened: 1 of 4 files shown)


JeremyKennedy (changes requested):
to investigate if this can be resolved without disabling server-side rendering

JeremyKennedy (approved):
I tried to fix this "properly" with SSR, but sadly nothing worked. Let's just merge this workaround in.

PR #347, opened 2022-07-15
Date inputs persistence

alex-solo:
### Description

List of proposed changes:

-
-


JeremyKennedy (approved):
I've tweaked your change to the Design System: https://github.com/DTS-STN/Service-Canada-Design-System/pull/340
This is because setting the `value` via code was preventing me from updating it myself by typing in the box, so I changed to using `defaultValue`. I've included the custom build in this PR - thanks Yarn Zero-Install!

I've also cleaned up your commit a bit, going back slightly to how it was before, but keeping the core changes (ie. useSessionStorage and passing month/year to the component.

PR #349, opened 2022-07-18
Fix bug where relying on the date input state would use out of date inputs

JeremyKennedy:
Recently caused by my own changes and lack of testing, this fixes an issue where the year input would use the previously-entered value, rather than the new value. Caused by a race condition / async isuse.


MarcoGoC (approved):
+1

PR #351, opened 2022-07-18
fix bug where last step would not properly check validity of form

JeremyKennedy:
A `return` in a `forEach` look doesn't act the same as a `return` in a `for` loop. Now the `return false` actually works as expected.

### DIFF
--- a/client-state/Form.ts
+++ b/client-state/Form.ts
@@ -93,9 +93,9 @@ export class Form {
   }
 
   get isValid(): boolean {
-    this.visibleFields.forEach((value) => {
-      if (!value.valid) return false
-    })
+    for (const field of this.visibleFields) {
+      if (!field.valid) return false
+    }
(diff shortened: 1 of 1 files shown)


MarcoGoC (approved):
uh

client-state/Form.ts
```diff
@@ -93,9 +93,9 @@ export class Form {
   }
 
   get isValid(): boolean {
-    this.visibleFields.forEach((value) => {
-      if (!value.valid) return false
-    })
+    for (const field of this.visibleFields) {
+      if (!field.valid) return false
```
MarcoGoC, line 97: doesn't it need a break here?

JeremyKennedy: line break? nope, this is totally valid syntax, I like it for simple stuff like this because it's concise!

PR #354, opened 2022-07-19
second half of "fix step validity"

JeremyKennedy:
I only submitted the first half of this fix in my previous PR #351


MarcoGoC (approved):
+1

PR #355, opened 2022-07-19
add more detail to "next steps" cards when income not provided

JeremyKennedy:
Previously we would only add "you can expect to receive X" when eligible - now we do it when income is missing as well.

Before:
image

After:
image


alexunique0519 (approved):
LGTM

PR #359, opened 2022-07-25
adding shortText for the answers and shortened the questions text

MarcoGoC:
### Description
List of proposed changes:
- Shorter questions Text
- Shorter answers Text by adding a new field called shortText on the TypedKeyAndText type

### Issues 
- Radio and Select uses a KeyAndText instead of TypedKeyAndText, so I changed the radio to use the 'new' type.
- This works on the browser but it wont compile it shows an error on YourAnswers.tsx line 77. I am not sure why still sees Radio with a KeyAndText type when I changed it to use TypedKeyAndText, I am missing one place to change it but I am not sure where.

- download ans run

### DIFF
--- a/components/ResultsPage/YourAnswers.tsx
+++ b/components/ResultsPage/YourAnswers.tsx
@@ -30,7 +30,6 @@ export const YourAnswers: React.VFC<{
           return (
             <div key={input.key} className="py-4 border-b-2 border-info-border">
               {tsln.resultsQuestions[input.key]} <br />
-              {console.log(input)}
               <strong>{getDisplayValue(input)}</strong> &nbsp;
               <DSLink
                 id={`edit-${input.key}`}
@@ -72,9 +71,10 @@ export const YourAnswers: React.VFC<{
             .text
         throw new Error(`values not found for field: ${input.key}`)
       case FieldType.RADIO:
-        if ('values' in fieldData)
+        if (fieldData.type === FieldType.RADIO && 'values' in fieldData) {
           return fieldData.values.find((value) => value.key === input.value)
             .shortText
+        }
         throw new Error(`values not found for field: ${input.key}`)
       default:
         throw new Error(`field type not supported in YourAnswers: ${fieldType}`)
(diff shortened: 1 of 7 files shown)


i18n/api/index.ts
```diff
@@ -24,6 +24,7 @@ export interface KeyAndText {
 export interface TypedKeyAndText<T> {
   key: T
   text: string
+  shortText: string
```
alex-solo, line 27: try making this optional: "shortText?"

components/ResultsPage/YourAnswers.tsx
```diff
@@ -66,11 +67,15 @@ export const YourAnswers: React.VFC<{
...
       case FieldType.DROPDOWN:
       case FieldType.DROPDOWN_SEARCHABLE:
-      case FieldType.RADIO:
         if ('values' in fieldData)
           return fieldData.values.find((value) => value.key === input.value)
             .text
         throw new Error(`values not found for field: ${input.key}`)
+      case FieldType.RADIO:
+        if ('values' in fieldData)
+          return fieldData.values.find((value) => value.key === input.value)
```
alex-solo, line 75: The problem seems to be that TS isn't sure if the field is going to be a dropdown or a radio. Check out the screenshot below:

Screen Shot 2022-07-26 at 9 54 04 PM

alex-solo: The solution is to convince TS that is in fact going to be a "FieldConfigRadio" when in this block. I would have expected TS to do that since we're in the "radio" case block but if I try doing this explicitly it seems to work ok:

Screen Shot 2022-07-26 at 9 57 13 PM

It doesn't seem like THE solution but perhaps it puts you on the right track? Interestingly enough, it doesn't work if I compare the declared variable "fieldType" to check if it's "radio". It has to be "fieldData.type"

alex-solo: @MarcoGoC not sure if you get notified when I leave comments but pinging you to check out the above. Let me know what you think or what you find.

MarcoGoC: Thanks Alex, I just saw the comments.

alex-solo (approved):
ok that should work!

PR #360, opened 2022-07-25
feat: dynamically generate links with state and benefit results

alexunique0519:
### Description

This PR is to fix the issue that the links after the summary box are not showing the correct link text. 

List of proposed changes:
- the first link shows not eligible or eligible text based on the state
- the second link shows estimate amount if eligible, otherwise not showing
- the fourth link shows the next steps for benefits if users are eligible for any, otherwise not show 
- last link shows the benefit that a user is not eligible for in our case, otherwise not show

### DIFF
--- a/components/ResultsPage/index.tsx
+++ b/components/ResultsPage/index.tsx
@@ -25,14 +25,12 @@ const notEligibleState = [
 
 // get the link text by current summary state
 const getEligibleLinkText = (
-  currentState: SummaryState,
+  entitlementSum: number,
   tsln: WebTranslations
 ): string => {
-  if (notEligibleState.includes(currentState)) {
-    return tsln.resultsPage.youAreNotEligible
-  } else {
-    return tsln.resultsPage.youMayBeEligible
-  }
+  return entitlementSum > 0
+    ? tsln.resultsPage.youMayBeEligible
+    : tsln.resultsPage.youAreNotEligible
 }
 
 const getEstimatedMonthlyTotalLinkText = (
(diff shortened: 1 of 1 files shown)


components/ResultsPage/index.tsx
```diff
@@ -3,18 +3,75 @@ import { useRouter } from 'next/router'
...
 import { FieldInput } from '../../client-state/InputHelper'
 import { WebTranslations } from '../../i18n/web'
-import { ResultKey } from '../../utils/api/definitions/enums'
+import { ResultKey, SummaryState } from '../../utils/api/definitions/enums'
 import {
   BenefitResult,
   BenefitResultsObject,
   SummaryObject,
 } from '../../utils/api/definitions/types'
+import Tbl1SingleScraper from '../../utils/api/scrapers/tbl1SingleScraper'
```
MarcoGoC: not used

components/ResultsPage/index.tsx
```diff
@@ -3,18 +3,75 @@ import { useRouter } from 'next/router'
...
 import { EstimatedTotal } from './EstimatedTotal'
 import { ListLinks } from './ListLinks'
 import { MayBeEligible } from './MayBeEligible'
 import { YourAnswers } from './YourAnswers'
+import { numberToStringCurrency } from '../../i18n/api'
+
+const notEligibleState = [
+  SummaryState.AVAILABLE_INELIGIBLE,
+  SummaryState.MORE_INFO,
+  SummaryState.AVAILABLE_INELIGIBLE,
```
MarcoGoC: duplicated value

components/ResultsPage/index.tsx
```diff
@@ -3,18 +3,75 @@ import { useRouter } from 'next/router'
...
+    }
+  }
+  return ''
+}
+
+const getBenefitsYouMayNotEligibleForLinkText = (
+  results: BenefitResultsObject,
+  tsln: WebTranslations
+): string => {
+  for (const key in results) {
```
MarcoGoC: consider passing resultsArray as parameter and using filter instead of for loop

components/ResultsPage/index.tsx
```diff
@@ -15,6 +15,62 @@ import { EstimatedTotal } from './EstimatedTotal'
...
+  SummaryState.MORE_INFO,
+  SummaryState.UNAVAILABLE,
+]
+
+// get the link text by current summary state
+const getEligibleLinkText = (
+  currentState: SummaryState,
+  tsln: WebTranslations
+): string => {
+  if (notEligibleState.includes(currentState)) {
```
MarcoGoC: you could use entitlementSum also

components/ResultsPage/index.tsx
```diff
@@ -15,6 +15,62 @@ import { EstimatedTotal } from './EstimatedTotal'
...
+    )}`
+  }
+  return ''
+}
+
+const getNextStepLinkText = (
+  results: BenefitResultsObject,
+  tsln: WebTranslations
+): string => {
+  for (const key in results) {
```
MarcoGoC: consider passing resultsArray as parameter and using filter instead of for loop

MarcoGoC (approved):
it works but may require some changes

components/ResultsPage/index.tsx
```diff
@@ -15,6 +15,60 @@ import { EstimatedTotal } from './EstimatedTotal'
 import { ListLinks } from './ListLinks'
 import { MayBeEligible } from './MayBeEligible'
 import { YourAnswers } from './YourAnswers'
+import { numberToStringCurrency } from '../../i18n/api'
+
+const notEligibleState = [
```
MarcoGoC: no longer used, otherwise all good

MarcoGoC (approved):
just remove the not used variables

PR #364, opened 2022-07-28
99999 gis bug part 2

MarcoGoC:
### Description

This change fixes cardDetail object to have the correct value when entitlement is **none** it also fixes the autoEnrollment value.  

### List of proposed changes:

- benefitHandler contains a few lines that overwrite some values but not all like mainText and autoEnrollment, those lines were copied to the _base benefit class.

- An over the limit combined income creates the scenario where mainText says is 'eligible' when it should not be.
- 
image


alexunique0519 (approved):
LGTM

PR #365, opened 2022-07-29
Fixed the inconsistent data picker styles

alexunique0519:
### Description
We imported the date picker component to our app, however, tailwindcss somehow overwrites the component's style with some default styles, and the date picker looks broken in our app. 

List of proposed changes:
- Fixed the height of the month field, now month and year look the same
- change the border color of the year field
- use rounded border corner for year field 
- remove the default chevron in month field, now the arrow looks good
- 
Screen Shot 2022-07-29 at 10 41 55 AM


MarcoGoC (approved):
It looks good on firefox

PR #373, opened 2022-08-18
Fixed the issue that radio options are not focusable by keyboard.

alexunique0519:
### Description

Problem: Radio options were not accessible through keyboard, and users were not able to enter the radio selection with keyboard.  

List of proposed changes:
- Radio options are focusable through Tab key and current focused option are enclosed with black rectangle. 
- Users are able to submit the section by pressing 'Enter' key

Pull the code and try it


Maxim-Lam (approved):
The functionality works great! might want to check in WET if there is a common styling used in highlighted fields.

PR #375, opened 2022-08-23
Added id for Select component.

alexunique0519:
### Description
This change is related to the issue that some of the inputs and select element in our accordion form don't have associated labels. 

List of proposed changes:
- add id for livingCountry select
- enable to expand/collapse tooltip with keyboard 

Inspect this element and check


alex-solo (approved):
tested locally, works as expected!

PR #378, opened 2022-08-24
Just for testing teamcity

MarcoGoC:
Testing team city

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -51,7 +51,7 @@
 
 # CMD [ "yarn", "start" ]
 
-FROM node:16.15.1 AS production
+FROM node:16.15.1-alpine AS production
 ENV NODE_ENV=production
 SHELL ["/bin/sh", "-c"]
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -69,7 +69,7 @@ RUN adduser \
 ENV NODE_ENV=production
 WORKDIR $home
 COPY --chown=55:$group . . 
-RUN yarn install --immutable
+RUN yarn install
```
alexunique0519: Do we need to change this or just leave it as it was?

pages/results/index.tsx
```diff
@@ -24,6 +24,8 @@ const ResultsPage = dynamic(
   { ssr: false }
 )
 
+// just for testing docker on teamcity
```
alexunique0519: we can remove this comment now.

MarcoGoC: I was not planning on merging this change

PR #380, opened 2022-09-01
Make text alts null for images that AT needs to ignore

alex-solo:
### Description

Refers to Bug [70440](https://dev.azure.com/VP-BD/DECD/_workitems/edit/70440) 

Summary of W3C F38 failure: All images that need to be ignored (such as icons) by AT (assistive technologies) should have null text alts so AT can ignore. 

List of proposed changes:

-
-


MarcoGoC (approved):
thanks

PR #381, opened 2022-09-02
ADO-70446

alexunique0519:
### Description

‘What you told us’ section on the right side panel has 7 links with the same text ‘Edit’, and clicking on each link takes the user back to the particular section in the previous page. 

When screen reader user navigates through these links it just reads ‘Edit’ for all links. Repeating link text like these can be confusing as it does not tell the user what to expect or where it goes or difference between each link. 

List of proposed changes:
- Use a customized anchor element to replace the DSLink component
- Add invisible supportive text to differentiate edit links when using a screen reader 

test with a screen reader, e.g. NVDA, chrome screen reader


MarcoGoC (commented):
I would say to use the title property instead and leave the text as is.

MarcoGoC (commented):
I agree with Alex we are supposed to use the DS components

alexunique0519:
close this PR as changes made in the design system

PR #382, opened 2022-09-02
Fix build issue in workflow

alexunique0519:
explicitly set node version to "16.15.1" to fix the build issue


MarcoGoC (approved):
good catch

PR #383, opened 2022-09-07
Added required suffix to all questions in the form

Maxim-Lam:
Added the "Required" suffix to all questions as per Sara's comments.

### DIFF
--- a/components/Forms/CurrencyField.tsx
+++ b/components/Forms/CurrencyField.tsx
@@ -3,14 +3,13 @@ import { InputHTMLAttributes, useEffect } from 'react'
 import NumberFormat from 'react-number-format'
 import { Language } from '../../utils/api/definitions/enums'
 import { ErrorLabel } from './validation/ErrorLabel'
-import { WebTranslations } from '../../i18n/web'
-import { useTranslation } from '../Hooks'
 
 export interface CurrencyFieldProps
   extends InputHTMLAttributes<HTMLInputElement> {
   name: string
   label: string
   helpText?: string
+  requiredText?: string
   error?: string
 }
 
@@ -27,11 +26,10 @@ export const CurrencyField: React.VFC<CurrencyFieldProps> = ({
   placeholder,
   onChange,
   helpText,
+  requiredText,
   error,
 }) => {
   const locale = useRouter().locale
-
-  const tsln = useTranslation<WebTranslations>()
 
   const localizedIncome =
     locale == Language.EN
@@ -60,8 +58,8 @@ export const CurrencyField: React.VFC<CurrencyFieldProps> = ({
         >
           {label}
         </label>
-        <span className="ds-inline ds-text-error-border-red ds-text-xl ds-font-medium pl-2">
-          ({tsln.required})
+        <span>
+          <span className="ml-1">({requiredText})</span>
         </span>
         {helpText && (
           <div className="ds-font-body ds-text-lg ds-leading-22px ds-font-medium ds-text-multi-neutrals-grey90a ds-mb-4">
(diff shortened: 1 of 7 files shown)


Maxim-Lam:
Ok guys, this is gonna be a little long so bare with me.

So I added the "required" suffix after each question (This is how it is in the figma file and Sara said just to follow the file)

Now the way that I added it is definitely not optimal but there is a reason. If you look in the MonthAndYear.tsx file, this is the proper way of appending the suffix but that is because it is using a Design System component that has the appropriate property to display the message; the other components are not from the Design System and cannot be handled that way.

I don't know if it would be beneficial or possible to either use Design System's component for all or it would be better to use none, either way it would be consistent.

If you guys have any suggestions I am all ears; as it stands, this is works but is sort of like a band-aid fix.

All feedback is appreciated :)

components/Forms/Radio.tsx
```diff
@@ -2,6 +2,7 @@ import { InputHTMLAttributes } from 'react'
 import { TypedKeyAndText } from '../../i18n/api'
 import { Tooltip } from '../Tooltip/tooltip'
 import { ErrorLabel } from './validation/ErrorLabel'
+import { FormLabel } from '@dts-stn/service-canada-design-system'
```
MarcoGoC: Added but not used?

components/Forms/Radio.tsx
```diff
@@ -43,6 +44,9 @@ export const Radio: React.VFC<InputProps> = ({
             className="mb-1.5 text-content font-bold question-link"
             dangerouslySetInnerHTML={{ __html: label }}
           />
+          <p className="ds-inline ds-text-error-border-red ds-text-xl ds-font-medium pl-2">
```
MarcoGoC: If inside the label tag maybe a span is better for this??

components/Forms/Radio.tsx
Maxim-Lam: Must have forgotten that there. I was playing around with it but there were issues. I will remove this

components/Forms/Radio.tsx
Maxim-Lam: Good idea.

components/Forms/CurrencyField.tsx
```diff
@@ -56,6 +56,9 @@ export const CurrencyField: React.VFC<CurrencyFieldProps> = ({
         >
           {label}
         </label>
+        <p className="ds-inline ds-text-error-border-red ds-text-xl ds-font-medium pl-2">
+          (Required)
```
alexunique0519: I guess 'Required' should be lowercase. Also, we may want to use an i18n variable here instead of hardcoding the label text here. It will still show 'Required' in French mode.

alexunique0519 (changes requested):

components/Forms/CurrencyField.tsx
Maxim-Lam: Good catch. I will make the change

alex-solo:
So far agree with Marco and Yan - lower case required in English and French using i8n and using span instead of p tag

components/Forms/Select.tsx
```diff
@@ -55,6 +59,9 @@ export const FormSelect: React.VFC<SelectProps> = ({
             {field.config.label}
           </span>
         </label>
+        <span className="ds-inline ds-text-error-border-red ds-text-xl ds-font-medium pl-2">
```
MarcoGoC: minor pick ds-inline is not required as span is an inline tag.

Maxim-Lam:
@MarcoGoC  I change the code to reflect how it was before it was removed

MarcoGoC (approved):
changes make sense to me, thanks

alexunique0519 (approved):

PR #386, opened 2022-09-09
70454 fix html

MarcoGoC:
### Description
Updating the html to avoid the a11y errors, 

**note** some of the errors are dependent on the DS components for which a PR have been submitted

List of proposed changes:
- removing `</br>` from the code and text
- removing duplicated id on the age question
- adding id attribute to radio group

- download and run


alexunique0519 (approved):
Looks good to me

PR #392, opened 2022-09-21
Added Aria-Label for Required suffix for screen readers to skip the P…

Maxim-Lam:
…arenthesis

### Description

Added Aria-label to "(Required)" suffix for screen readers to ignore the Left and right "paren"

Either select element with Screen reader or inspect element to see aria-label was added

### DIFF
--- a/components/Forms/CurrencyField.tsx
+++ b/components/Forms/CurrencyField.tsx
@@ -59,7 +59,9 @@ export const CurrencyField: React.VFC<CurrencyFieldProps> = ({
           {label}
         </label>
         <span>
-          <span className="ml-1">({requiredText})</span>
+          <span className="ml-1" aria-label={requiredText}>
+            ({requiredText})
+          </span>
         </span>
         {helpText && (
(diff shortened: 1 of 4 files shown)


MarcoGoC (commented):
by following example #4 of this page 
https://www.w3.org/TR/WCAG20-TECHS/ARIA2.html 
the required should be inside the `label` tag. and adding a `aria-required="true"` to the control. 

also according to this page https://a11y-101.com/development/required section `The word` 
`(required)` should be inside the label. 

I am guessing that is the reason why the screen reader is reading left paren - right paren

MarcoGoC (changes requested):
see comment

Maxim-Lam:
> Yes, I believe it should be <span> can help with the styling

I can certainly make those changes.

My only concern was that the Design System Component we are using for the date has the label text and required text in separate paragraphs and therefore separate styles. Adding the required text into the label with give it the same styles and make it less distinguishable.

@MarcoGoC That being said, do you recommend I still just add it as part of the label?

alexunique0519:
Close this PR since another PR for this fix has been merged.

PR #393, opened 2022-09-23
modifying id for each radio control

MarcoGoC:
### Description

List of proposed changes:
- modifying the id to avoid duplication

- run and validate page passes html validator  

### Additional Notes


alexunique0519 (dismissed):
we could change id to {`radio-${name}-${index}`} because we will have duplicated {`radio-${index}`} in each card.

alexunique0519 (approved):

PR #396, opened 2022-09-27
Update aria labels in "What you told us"

alexunique0519:
### Description

The "edit" links aria labels have been updated according to the suggested wordings. Both En and Fr. 

pull code and inspect or test with screen readers


components/ResultsPage/YourAnswers.tsx
```diff
@@ -37,9 +37,7 @@ export const YourAnswers: React.VFC<{
                   href={`eligibility#${input.key}`}
                   text={tsln.resultsPage.edit}
                   target="_self"
-                  ariaLabel={`${tsln.resultsPage.edit} ${
-                    tsln.resultsQuestions[input.key]
-                  }`}
+                  ariaLabel={tsln.resultsPage.yourAnswers[input.key]}
```
MarcoGoC, line 40: I think what Jeremy wanted us to use is
` tsln.resultsQuestions[input.key] ` 
I think you should add the new text to the resultsQuestions object instead

MarcoGoC (changes requested):
see comment thanks

components/ResultsPage/YourAnswers.tsx
alexunique0519: In en.ts, we can find "resultsQuestions: **apiEn.questionShortText**", therefore if we want to still use resultsQuestions, we have to add new translations into **questionShortText**. My concern is that is it good to do so? Because what we are going to add are not questions nor short version of questions, we are just adding aria labels.

PR #401, opened 2022-10-07
Remove the redundant word "icon"

alexunique0519:
### Description

According to the accessibility recommendation on Miro, change the alt text of warning image from "warning icon" to "warning".  Also added the text to 118n translations instead of using the hardcoded word. 

Pull the code to inspect the alt text in both Fr and En mode and test with VoiceOver or NVDA

### Additional Notes
N/A


MarcoGoC (approved):
Thanks

PR #403, opened 2022-10-14
fix: fixed the accessibility issues with radios

alexunique0519:
### Description

This pr includes a series of fixes regarding the accessibility issue with radios. 

List of proposed changes:
- Fixed the issue that the screen reader could not recognize a radio. 
- Fixed the issue that the screen reader could not tell how many options in each group.
- Fixed the issue that the screen reader could not tell which options is selected. 
- Fixed the issue that the label text was read twice for each option. 
- Fixed the issue that the Up/Down key can not select options

Pull the changes and test with VoiceOver / NVDA


MarcoGoC:
branch name is not TC-Staging but a brand new branch staging

PR #415, opened 2022-10-26
change radio label cursor to be pointer when hover over

alexunique0519:
### Description
 Change the cursor type from default to pointer. In this way, when user hovers over the mouse cursor on radio labels, the cursor style change will indicate the label is clickable for selecting an option.

pull code and test it on browser 

### Additional Notes

N/A


alex-solo (approved):
LGTM!

PR #426, opened 2022-11-01
updated currencyField and Income Text

MarcoGoC:
### Description

Update Questions page: Income and add icon with "textbox" 

- Income text must be the new one. 
- click on the info icon to display the extra


alexunique0519 (dismissed):
Everything looks good except small style difference in the tooltip text format. For example, "You can find..." and "For a more accurate" should have a space/margin in between. Also, there is no indentation for the bullets.

alex-solo:
Screen Shot 2022-11-02 at 11 30 30 AM

There is also this weird ::after element that is clickable which shouldn't be there

alex-solo (dismissed):
See comment

MarcoGoC:
> Everything looks good except small style difference in the tooltip text format. For example, "You can find..." and "For a more accurate" should have a space/margin in between. Also, there is no indentation for the bullets.

updated

MarcoGoC:
> Screen Shot 2022-11-02 at 11 30 30 AM
> 
> There is also this weird ::after element that is clickable which shouldn't be there

updated, tailwind sometimes takes css from the DS and sometimes does not it's a pain

alexunique0519:
Screen Shot 2022-11-03 at 2 13 09 PM
I think the text area should stay under the icon, it currently breaks the mobile view.

PR #428, opened 2022-11-02
ADO 83990: Made changes to the residence history question card and fixed label in age card

alex-solo:
### Description

Refer to ADO task 83990. In addition, I noticed that the "required" text in the age card wasn't toggling to French when language was changed. I will bring this up to the design system team but luckily, the component is flexible enough that we can just turn off their label and use our own. In a separate PR, I will implement the label we use in each question as its own component.

List of proposed changes:

-
-

### DIFF
--- a/components/EligibilityPage/index.tsx
+++ b/components/EligibilityPage/index.tsx
@@ -79,7 +79,6 @@ export const EligibilityPage: React.VFC = ({}) => {
 
     if (button) {
       button.setAttribute('type', 'submit')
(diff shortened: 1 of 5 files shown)


MarcoGoC (commented):
not sure if is the image or they are actually asking for a less wide box for text. 
...outside of Canada... is two separate lines

alex-solo:
@MarcoGoC No, I wouldn't worry about that.

MarcoGoC (approved):
Ok, I think Yan did made a change don't remember the task to be w-75%

PR #429, opened 2022-11-02
Add label component so there is no code duplication

alex-solo:
### Description

To avoid code duplication, extract label code into its own component

List of proposed changes:

-
-

Check that all labels are exactly as they were before and match Figma designs

### DIFF
--- a/components/Forms/NumberField.tsx
+++ b/components/Forms/NumberField.tsx
@@ -50,7 +50,7 @@ export const NumberField: React.VFC<NumberFieldProps> = ({
       />
       {error && <ErrorLabel errorMessage={error} />}
       <NumberFormat
-        id={name}
+        id={`enter-${name}`}
         name={name}
         className={`form-control text-content border-form-border ${
           error ? ' border-danger' : ''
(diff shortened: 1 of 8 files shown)


alexunique0519 (approved):
LGTM

components/Forms/QuestionLabel.tsx
```diff
@@ -20,7 +20,7 @@ export const QuestionLabel: React.FC<QuestionLabelProps> = ({
       <label
         htmlFor={name}
         aria-label={name}
-        data-testid={`${type}-input-label`}
+        data-testid={`${type}-label`}
```
MarcoGoC, line 23: the input or any other name is required because of id duplication the name of the card and the input is the same

components/Forms/NumberField.tsx
```diff
@@ -43,7 +43,7 @@ export const NumberField: React.VFC<NumberFieldProps> = ({
     <>
       <QuestionLabel
         name={name}
-        type="number"
+        type="number-input"
```
MarcoGoC, line 46: the input or any other name is required because of id duplication the name of the card and the input is the same

MarcoGoC (changes requested):
See comments

MarcoGoC (approved):

components/Forms/NumberField.tsx
alex-solo: Good find Marco. The number input id was the same as the accordion form card id. I just pushed a change

MarcoGoC (approved):
looks good

PR #435, opened 2022-11-10
updated the validation schema and translations based on 84583

alexunique0519:
### Description

Updated the validation schema and messages translations according to 84583. For details, please refer to task descriptions. 

Pull changes and run


utils/api/definitions/schemas.ts
```diff
@@ -61,35 +61,43 @@ export const RequestSchema = Joi.object({
...
-          'any.unknown': ValidationErrors.yearsInCanadaNotEnough,
-        }),
-      }),
+    .when('livingCountry', {
+      is: Joi.string().valid('CAN'),
+      then: Joi.when(
+        'yearsInCanadaSince18',
+        //9.9 to disable 10,since the input only accepts integer, this is good
+        {
+          is: Joi.number().max(9.9),
```
alex-solo: Is it possible to use 10 and 20 in validating the input and not 9.9 and 19.9?

i18n/web/en.ts
```diff
@@ -156,8 +156,11 @@ const en: WebTranslations = {
     [ValidationErrors.partnerAgeOver150]:
       "Your partner's age should be less than 150.",
     [ValidationErrors.oasAge65to70]: 'You must enter an age between 65 and 70.',
-    [ValidationErrors.yearsInCanadaNotEnough]:
-      'The number of years you have lived in Canada is not enough to receive any old age benefits.',
+    [ValidationErrors.yearsInCanadaNotEnough10]:
+      //'The number of years you have lived in Canada is not enough to receive any old age benefits.',
```
alex-solo: Why are some of these commented out?

alex-solo (changes requested):
See comments

utils/api/definitions/schemas.ts
alexunique0519: if we use max(10) here, it mean [0, 10], which includes 10 years. But we only need to show the error when user lives under 10 years. So we can't use max(10)

i18n/web/en.ts
alexunique0519: The error messages have been changed, I will remove the commented messages.

utils/api/definitions/schemas.ts
alex-solo: I mean using a different Joi method (something like "under" or "lessThan" if those exist) to work in terms of numbers that we care about (10 and 20)

alexunique0519: replace the max() with less(), works like a charm

alex-solo (approved):
issues resolved

PR #436, opened 2022-11-10
83697 Income text updates as per task

MarcoGoC:
### Description

List of proposed changes:
- text changes 

- Download and verify


alexunique0519 (dismissed):
I think the text should be inside a "More information" tooltip like the one in Legal status card.

alex-solo (approved):
LGTM

PR #437, opened 2022-11-14
change radio button cursor style

alexunique0519:
### Description

Changed the cursor style when hovering over radio buttons to be pointer.


MarcoGoC (commented):
to be determined if this is the expected behavior, as the DS radio acts as a single 'element' as opposed to ours 2 elements the icon breaks (or goes back to normal) when hover horizontally in-between the radio and the label.

MarcoGoC (approved):
that should do it,  the hover is no longer necessary on the input and label though

PR #439, opened 2022-11-15
85733 updated OAS results text

MarcoGoC:
### Description

List of proposed changes:
- text changes 

- Download and verify


alex-solo (changes requested):
As per note on the ADO task (from Sara) we are going ahead with continuing to use the "More information" drop down.

I spoke with Andreanne and she confirmed that we can just use the "More information" dropdown as before to complete this task. 

So wherever you see a blue icon, just add "More information" dropdown and the content changes shown in the blue boxes in the ADO task.

alex-solo (approved):

PR #454, opened 2022-12-01
Make change to GIS calculation for above 75 client

alex-solo:
Addresses bugs: ADO-913, 91317

merge to match develop## [DC-XXX](https://jira-dev.bdm-dev.dts-stn.com/browse/DC-XXX) (Jira Issue)

### Description

Refer to the bugs in ADO for description of the problem.

Solution:
When a client is 75, in the results object for OAS we don't keep track of what the OAS amount would have been from ages 65-74 but we need that amount to adjust the GIS calculation. We were previously subtracting OAS "result" which refers to the current OAS amount ($ amount the client would get given their age in the form). So even when a client is above 75, we should keep track of the original OAS amount (not increased by 10% due to age) to use in the GIS calculation. 

List of proposed changes:

-
-


MarcoGoC (approved):
lgtm

PR #464, opened 2022-12-09
90492 Fous on the new element

MarcoGoC:
### Description
When user selects an option for "Current marital status", it reads that it is selected, then maybe it should take the user right down to "Are you able to provide us your partner's annual net income?", to let them know this has opened up?

List of proposed changes:
- Added autoFocus to the fields, not exactly sure if they meant the focus on the label or the actual control, will see how it works on a screen reader. 

- Verify on the [dynamic url](https://eligibility-estimator-dyna-90492-autofocus.bdm-dev.dts-stn.com/)

### DIFF
--- a/components/Forms/CurrencyField.tsx
+++ b/components/Forms/CurrencyField.tsx
@@ -76,6 +76,7 @@ export const CurrencyField: React.VFC<CurrencyFieldProps> = ({
         placeholder={placeholder}
         onChange={onChange}
         required
+        autoFocus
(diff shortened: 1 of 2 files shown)


alexunique0519 (changes requested):
I tested with the latest changes. When I select "Married or common-law", the screen reader didn't bring the focus to the next question. I assume this is what the task requires.

MarcoGoC:
wont be done

PR #466, opened 2022-12-13
add jquery to EE to test AA form tracking

alexunique0519:
### Description

Add jquery to the <head> section as AA requires jquery.

### DIFF
--- a/pages/_document.tsx
+++ b/pages/_document.tsx
@@ -10,6 +10,11 @@ export default function Document() {
           href="https://fonts.googleapis.com/css2?family=Lato:wght@400;700&family=Noto+Sans:ital,wght@0,400;0,700;1,400;1,700&display=swap"
           rel="stylesheet"
         />
+        <script
+          src="https://code.jquery.com/jquery-3.6.2.min.js"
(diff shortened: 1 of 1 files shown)


pages/_document.tsx
```diff
@@ -10,6 +10,11 @@ export default function Document() {
           href="https://fonts.googleapis.com/css2?family=Lato:wght@400;700&family=Noto+Sans:ital,wght@0,400;0,700;1,400;1,700&display=swap"
           rel="stylesheet"
         />
+        <script
+          src="https://code.jquery.com/jquery-3.6.2.min.js"
```
alex-solo, line 14: Are you thinking of using CDN just to test things or as a permanent solution? We should probably include jquery as a dependency and install it instead of using CDN for our actual implementation. Let me know what you think.

alex-solo (changes requested):

pages/_document.tsx
alexunique0519: let's try it first

alex-solo: decided to approve so we can test. Will pull in dependency if this works so we have a self contained app for production

alex-solo (approved):

PR #469, opened 2022-12-14
ADO-93507: Removed income constraints so testing can resume

alex-solo:
### Description

Commented out income checks for now so testing can resume. We will likely need these constraints in Joi when implementing the updated error validation. 

List of proposed changes:

-
-

When the income for individual or combined income with partner reaches the upper limit, user should still be able to proceed with filling out the form.


alexunique0519 (approved):
maybe we can change some of the unit tests later, for now its good.

PR #474, opened 2022-12-15
ADO-92499: Fix the age constraint

alex-solo:
### Description

We only need to restrain the lower bound on age. If the age is over 60, then display the legal status questions for partner. Because they might be eligible not just for ALW but other benefits as well

List of proposed changes:

-
-


MarcoGoC (approved):
looks good

PR #476, opened 2022-12-16
Aa changes

alexunique0519:
This PR is to address the recommendations from Andrew regarding the AA part in our code. So the form data collection is removed, for example,  'data-gc-analytics-collect' attribute.

Also we add  'data-gc-analytics-customclick' attribute for all next step button, add 'data-gc-analytics-formsubmit='submit' 
' for 'estimate my result' button


MarcoGoC (approved):
LGTM :)

PR #477, opened 2022-12-19
ADO 90955 - Fix the ability to handle input change of questions before an input that has an error

alex-solo:
### Description

When one of the inputs has an error, the user was not able to make changes to other inputs. For example whether a question is visible is dependant on the response to a previous question and the latter response produces an error, we weren't able make the latter question disappear if the previous response was altered. 

The inputs were indeed changing (input object in dev tools for example) but we were not changing the "visibleFields" object in the error case during Joi validation. 

Before: 
User selects "I would like to delay my first payment" (oasDefer: true) then enters an invalid value for the input that opens up (oasAge). Then change the option for the previous question to (oasDefer: false). The field oasAge should go away but before this fix, it did not. 

Try other scenarios that produce errors and change inputs for other questions


MarcoGoC (approved):
not exactly sure why a new object is required

PR #480, opened 2022-12-21
feat: send the button attributes as props instead of use UseEffect

alexunique0519:
### Description
This PR is to address the issue that adding the attributes for the accordionForm buttons with useEffect. I have made change to accordion form to accept props of buttonAttributes so that we can pass those attributes from EE app directly without using useEffect hook. 

run the app and inspect button attributes.


components/EligibilityPage/index.tsx
```diff
@@ -128,26 +85,46 @@ export const EligibilityPage: React.VFC = ({}) => {
       title: tsln.category.age,
       buttonLabel: `${tsln.nextStep}${connection} ${tsln.category.income}`,
       keys: getKeysByCategory(FieldCategory.AGE),
+      buttonAttributes: {
+        'data-gc-analytics-customclick':
+          'ESDC-EDSC:Canadian OAS Benefits Est. Next Step Click:Income',
```
alex-solo: Can we make these dynamic from tsln in case the titles of cards change?

alex-solo (changes requested):

MarcoGoC (approved):
works ok for me

alex-solo (approved):

PR #481, opened 2022-12-21
fix: fixed the issues in ado-93713

alexunique0519:
### Description

This pr is to fix all related issues in ado-93713. Some labels of the input or select fields in the form don't have a  htmlFor props, because it uses the {name} to be id, which is duplicated with the section id.  The fix is to differentiate the label's forHtml prop with the section div's Id. 

Run the app and inspect the element


alex-solo (approved):
Doesn't affect functionality as far as I could tell during a quick fetch and test. All good

PR #487, opened 2022-12-22
ADO 87406 - Add error messages

alex-solo:
… given step## [DC-XXX](https://jira-dev.bdm-dev.dts-stn.com/browse/DC-XXX) (Jira Issue)

### Description

Please refer to link to Miro provided in ADO task 87406 to see the list of changes and correct text that should be seen for each error.

Pull, compare in English and French to the Miro designs. Test behaviour by entering incorrect ranges, leaving inputs empty, etc.


MarcoGoC (approved):
works ok 
only issue i have found is that, when going back an changing from income provided equal no to yes an error is displayed automatically for the amount 

for the record 
I do not like the new design as an error is supposed to display a box, which takes the focus away from the button.

alex-solo:
> 

Thanks for the feedback. Yeah, it's not perfect. Once the "next" button is pressed then the errors are revealed. Unfortunately, the way the accordion form works, I cannot register a "Next" click again on the same card if the subsequent card is now open and in focus. This is also a problem in the reverse case; once you move on from a card and a non-visible question's error visibility is set to false, that error won't show up once I go back to a previous card and toggle a field to visible. For example revealing "oasAge" field when setting "oasDefer" to true. 

Anyway, some bugs to work out but will require changing the behaviour of the accordion form "Next" buttons. Will get Andreanne to start testing this to see what's acceptable for now.

PR #488, opened 2022-12-23
Design changes that make no sense, imo

MarcoGoC:
### Description
  another change requested to remove 'delay' and 'income provided', and display only the age or the income
 
List of proposed changes:
- as above. 

- Verify on the [dynamic url](https://eligibility-estimator-dyna-85662-youranswers.bdm-dev.dts-stn.com/)

### Additional Notes
- this could be avoided by changing the questions instead.

### DIFF
--- a/components/ResultsPage/YourAnswers.tsx
+++ b/components/ResultsPage/YourAnswers.tsx
@@ -18,6 +18,28 @@ export const YourAnswers: React.VFC<{
   )
 
   /**
+   * shouldDisplay
+   *    Returns  False  when Income Provided is Yes or OAS deferral is Delay
+   *    Returns  True   for any other scenario
+   */
+  function shouldDisplay(input: FieldInput): boolean {
+    let returnVal: boolean
+
+    input.key !== 'incomeAvailable' &&
+    input.key !== 'partnerIncomeAvailable' &&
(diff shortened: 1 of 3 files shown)


components/ResultsPage/YourAnswers.tsx
```diff
@@ -17,6 +17,28 @@ export const YourAnswers: React.VFC<{
     tsln._language
   )
 
+  /**
+   * shouldDisplay
+   *    Returns  False  when Income Provided is Yes or OAS deferral is Delay
+   *    Returns  True   for any other scenario
+   */
+  function shouldDisplay(input: FieldInput): boolean {
```
alex-solo, line 25: Seems to work as expected but may I suggest rewriting the "shouldDisplay" function? Currently it's a bit hard to read and I don't think you need the first check? I tried this and it works as well:

```
function shouldDisplay(input: FieldInput): boolean {
    const dontShow: String[] = ['incomeAvailable', 'partnerIncomeAvailable', 'oasDefer']
    return !dontShow.includes(input.key) && input.value !== 'true'
 }
```

alex-solo (dismissed):

alex-solo (approved):
lgtm

PR #490, opened 2023-01-05
Add AA button click even to Start Estimate button

alexunique0519:
This change enables AA to send requests when users click the "start benefit estimator" .

Need to deploy and test.


MarcoGoC (approved):
make sense

PR #495, opened 2023-01-09
93663 changing alws formula

MarcoGoC:
### Description
  Changing the formula to be less than 65. as the age is calculated value with a decimal   
 
List of proposed changes:
- as above. 

- Verify on the [dynamic url](https://eligibility-estimator-dyna-93663-alws.bdm-dev.dts-stn.com/)

### Additional Notes
-


alexunique0519 (approved):
LGTM

PR #502, opened 2023-01-16
ADO 96375 - fix bug that shows January when going back to eligibility page

alex-solo:
### Description

Check out ADO 96375 for details

List of proposed changes:

- use useEffect hook to set date inputs as local state in the MonthYear component. 

- Pull changes
- change month to something other than January
- From the results page, click on "Edit Answers" and check that the month is still the month you have selected
- From the results page, click on "Edit" in the "What you told us" section and check that the month is still the month you have selected

### DIFF
--- a/components/Forms/MonthAndYear.tsx
+++ b/components/Forms/MonthAndYear.tsx
@@ -1,3 +1,4 @@
+import { useState, useEffect } from 'react'
 import { DatePicker, FormError } from '@dts-stn/service-canada-design-system'
 import { debounce } from 'lodash'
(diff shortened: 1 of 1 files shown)


MarcoGoC (approved):
Interesting that for the partners age the control works correctly but not for the main applicant I thought it was a name mismatch somewhere

alex-solo:
> Interesting that for the partners age the control works correctly but not for the main applicant I thought it was a name mismatch somewhere

Yeah, that's odd actually. You'd expect it to work the exact same way. Only difference in how deep the DatePicker is nested in the case of the partner, I'm assuming that is why.

PR #504, opened 2023-01-17
Fix radio issues

alexunique0519:
### Description

Fixed the radio issues and some HTML issues following the w3 validator (nu checker)

### DIFF
--- a/components/Forms/Radio.tsx
+++ b/components/Forms/Radio.tsx
@@ -44,33 +44,34 @@ export const Radio: React.VFC<InputProps> = ({
           helpText={helpText}
         />
         <Tooltip field={name} />
-        {values.map((val, index) => (
-          <div
-            role="radiogroup"
-            key={index}
-            id={`${name}-r${index}`}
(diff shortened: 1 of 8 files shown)


components/Forms/Radio.tsx
```diff
@@ -36,54 +36,47 @@ export const Radio: React.VFC<InputProps> = ({
...
       <fieldset>
-        {/* <QuestionLabel
+        <QuestionLabel
           name={name}
           type="radio"
           label={label}
           requiredText={requiredText}
           helpText={helpText}
-        /> */}
-        <legend
```
MarcoGoC, line 46: not sure IF not having a legend would pass a11y

alexunique0519: I tried on the nu checker, it didn't complain

alexunique0519: @MarcoGoC  we already have a legend for each card, I don't think we need legend for each radio

alex-solo:
@alexunique0519 
image

Still seeing the styling that we don't want. Is that something you don't see when testing it?

alex-solo (approved):
seems to be working. Thanks for pushing this along!

PR #506, opened 2023-01-17
Fix alert and form errors

alex-solo:
### Description

Some form errors were showing up as alert errors. 

Specifically, we should check "legalStatus" and "everLivedSocialCountry". These questions should still produce a form error (red border and message under the input) when empty. However, when a selection is made, in the case of legalStatus, "Other" should produce an alert message. If an error is an alert, it should not appear in the list of errors on the top of the page.


MarcoGoC:
Playing around and not following a forward pattern has the issue of not getting an error when leaving an empty field. is not the idea of how it should be used normally but still

alex-solo:
> Playing around and not following a forward pattern has the issue of not getting an error when leaving an empty field. is not the idea of how it should be used normally but still

When left empty on the "legalStatus" question, you should get an error. Is that not what you see?

alexunique0519 (approved):
change works for me

PR #509, opened 2023-01-19
Fix error validation

alex-solo:
### Description

This is related to the bug that Yan noticed - when going back to change the birth year, the error no longer appeared. It is because I missed another clause in the if statement to determine whether a form error should be shown.

List of proposed changes:

-
-


alexunique0519 (approved):
tested, works for me

PR #511, opened 2023-01-20
Rewrite everLivedSocialCountry error logic using custom function in Joi

alex-solo:
### Description

`everLivedSocialCountry` did not throw an error when no selection was made. The regular Joi.required().messages() pattern did not work here because of the "when" function in Joi. 

List of proposed changes:

- use "custom" function and reference other entries with `helpers.state` to do the comparisons

Should function as before when selecting Yes/No but should also give an error when no selection is made


MarcoGoC (approved):
looks good to me 
wait until to merge until the latest helm change is build to test staging works as it used to

PR #513, opened 2023-01-23
Changed debounce to higher interval to fix the input bounce problem

alex-solo:
### Description

Tiny change to fix the issue of the cursor bouncing left and right after every input change.

In the currency field, the cursor should no longer move left or right and maintain its position at the last entered or deleted value.

### DIFF
--- a/components/EligibilityPage/index.tsx
+++ b/components/EligibilityPage/index.tsx
@@ -190,14 +190,14 @@ export const EligibilityPage: React.VFC = ({}) => {
       keyStepMap[step].keys.includes(key)
     )
 
-    if (nextForStepClicked[step]) {
-      setErrorsVisible({ ...errorsVisible, ...getVisisbleErrorsForStep(step) })
-    }
-
     field.value = newValue
     inputHelper.setInputByKey(field.key, newValue)
     form.update(inputHelper)
     setCardsValid(getStepValidity())
+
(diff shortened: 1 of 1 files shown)


components/EligibilityPage/index.tsx
```diff
@@ -261,7 +261,7 @@ export const EligibilityPage: React.VFC = ({}) => {
                 label={field.config.label}
                 onChange={debounce(
                   (e) => handleOnChange(field, e.target.value),
-                  100
+                  500
```
MarcoGoC: although it helps, still has issues when entering the income very slow, not sure why is this happening as older branches didn't have this issue

MarcoGoC (changes requested):
not sure why is this happening

components/EligibilityPage/index.tsx
alex-solo: hmm, I tested again and it seems to be working fine on initial input but when going back to the input, the bouncing back and forth comes back. Darn.. I'm not sure what caused this to start happening but I will investigate a bit more.

alex-solo: Ah, I simply needed to move the setState for error visibility after React actually deals with changing the input state!

MarcoGoC (approved):
works as before +1

PR #521, opened 2023-01-26
ADO90305 - show partner eligibility on results page

alex-solo:
### Description

If the client chooses the options "I don't know" or "No, my partner does not receive the Old Age Security pension", we calculate the partner eligibility and if the partner is eligible for any of the benefits (OAS, GIS, ALW), then the client will receive the message that "Your partner may be eligible" and see how much they would be eligible for. 

For complete details on what things should look like please refer to the ADO task.

List of proposed changes:

- MainHandler to return allResults instead of allResults.client
- Pass down both `results` and `partnerResults` to props that require it (results page and the benefit cards)
- merge cardCollapsedText from partnerResults with results and display in benefit cards as before
- BONUS: fix shouldDisplay function to avoid nested tertiary statement and be more readable

If partner is eligible for OAS, GIS or ALW, the amount and message should appear in the associated BenefitCard on the results page.

### DIFF
--- a/components/ResultsPage/BenefitCards.tsx
+++ b/components/ResultsPage/BenefitCards.tsx
@@ -8,7 +8,8 @@ import { BenefitCard } from './BenefitCard'
 
 export const BenefitCards: React.VFC<{
   results: BenefitResult[]
-}> = ({ results }) => {
+  partnerResults: BenefitResult[]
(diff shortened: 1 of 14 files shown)


alexunique0519 (approved):
LGTM

alexunique0519 (approved):

PR #537, opened 2023-02-03
Added test site banner

alexunique0519:
### Description

Added Test Site banner as per request. 

[Test url](http://ep-be-dyna-add-testbanner.bdm-dev-rhp.dts-stn.com/)


MarcoGoC (approved):
looks good

PR #541, opened 2023-02-09
fix: check if the incomeIsProvided outside first.

alexunique0519:
### Description

Just realized the first condition check I had for yesterday's fix was not correct. The income-provide check should cover all scenarios. Therefore, move the check to the outer scope to cover all scenarios.


alex-solo (approved):
no breaking changes

PR #542, opened 2023-02-09
103540  updated formula alw

MarcoGoC:
### Description
Updating formula so that the partner is eligible 

List of proposed changes:
- as above. 

- Verify on the [dynamic url](https://ep-be-dyna-103540-alw.bdm-dev-rhp.dts-stn.com/)


alex-solo (approved):
Looks good, returns the correct amount for partner ALW

alexunique0519 (approved):

PR #543, opened 2023-02-10
103564 - updated GIS results when negative

MarcoGoC:
### Description
The main issue is the GIS amounts are negative and subtract from the total payment. 
  see console logs for debugging from Yan. 
The changes are so that the function returns 0 instead of a negative amount.  

Adding changes done to 103540 because of an issue with one unsigned commit. this would make it easier as opposed to fix the other branch.

List of proposed changes:
- as above. 

- Verify on the [dynamic url](https://ep-be-dyna-103564-alw.bdm-dev-rhp.dts-stn.com/)


alexunique0519 (approved):
LGTM

PR #545, opened 2023-02-10
Gis result changes

alexunique0519:
### Description

Implemented all UI changes and logic changes regarding the new design of the GIS result card. Please refer to the figma when reviewing this PR.


alex-solo (approved):
matches the "GIS result" section in Figma, except the very last scenario when "Not eligible".

PR #549, opened 2023-02-14
104215 - Avoid negative amounts

MarcoGoC:
### Description
The main issue is the GIS amounts are negative and subtract from the total payment. 

List of proposed changes:
- as above. 

- Verify on the [dynamic url](https://ep-be-dyna-104215-negativegis.bdm-dev-rhp.dts-stn.com/)


alexunique0519 (approved):
Works for me

PR #555, opened 2023-02-17
ADO 107231 identical ids

Maxim-Lam:
### Description

ids for the BenefitCards and overlapped with those of the "on this page" ListLinks.

Replaced ids for both with unique identifiable ids

List of proposed changes:

- replaced duplicate ids from benefitsCard and ListLinks to unique ones

- navigate to /eligibility
- fill out form with valid data
- inspect ID on the next page for the "on this page" links and the benefits cards "apply for [...]" and "learn more [...]"
- notice unique IDs

### DIFF
--- a/components/ResultsPage/BenefitCard.tsx
+++ b/components/ResultsPage/BenefitCard.tsx
@@ -17,7 +17,6 @@ export const BenefitCard: React.VFC<{
     url: string
     text: string
     alt: string
-    idSuffix: string
   }>
 }> = ({
   benefitKey,
@@ -77,13 +76,17 @@ export const BenefitCard: React.VFC<{
 
       <div className="mt-4">
         {links &&
-          links.map(({ text, url, icon, alt, idSuffix }, index) => (
+          links.map(({ text, url, icon, alt }, index) => (
             <div key={index} className="flex items-center py-4 text-content">
               <div>
                 <Image src={`/${icon}.png`} alt={alt} width="30" height="44" />
               </div>
               <div className="pl-5 w-full">
-                <DSLink id={`${idSuffix}Link${index}`} href={url} text={text} />
+                <DSLink
+                  id={`${benefitKey}Link${index}`}
+                  href={url}
+                  text={text}
+                />
               </div>
             </div>
           ))}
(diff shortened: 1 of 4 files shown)


components/ResultsPage/index.tsx
```diff
@@ -74,31 +74,42 @@ const ResultsPage: React.VFC<{
   let listLinks: {
     text: string
     url: string
+    idSuffix: string
   }[] = [
     {
       text: getEligibleLinkText(summary.entitlementSum, tsln),
       url: '#eligible',
+      idSuffix: 'eligilble',
```
MarcoGoC: typo

components/ResultsPage/BenefitCard.tsx
```diff
@@ -71,13 +77,13 @@ export const BenefitCard: React.VFC<{
...
         {links &&
-          links.map(({ text, url, icon, alt }, index) => (
+          links.map(({ text, url, icon, alt, idSuffix }, index) => (
             <div key={index} className="flex items-center py-4 text-content">
               <div>
                 <Image src={`/${icon}.png`} alt={alt} width="30" height="44" />
               </div>
               <div className="pl-5 w-full">
-                <DSLink id={`link${index}`} href={url} text={text} />
+                <DSLink id={`${idSuffix}Link${index}`} href={url} text={text} />
```
MarcoGoC: suggestion use Link${benefitKey}-${index}  instead, so idSuffix is not only not required, but no other change is required.

components/ResultsPage/index.tsx
```diff
@@ -74,31 +74,42 @@ const ResultsPage: React.VFC<{
   let listLinks: {
     text: string
     url: string
+    idSuffix: string
```
MarcoGoC: with the suggestion above this is not required

components/ResultsPage/BenefitCard.tsx
Maxim-Lam: good idea

components/ResultsPage/index.tsx
Maxim-Lam: @MarcoGoC  An ID suffix is in someway required for listLinks as unlike the Benefits card, there is no benefitKey passed down to them. I did remove them in lieu of the benefits key on the BenefitsCard component.

MarcoGoC (approved):
Thanks Max

PR #556, opened 2023-02-17
Fixed ado-107356 by generating card details with single inputs

alexunique0519:
## Ado-107356

This PR also includes the fix for ADO-104215 and ADO-107356 regarding the scenario when the applicant's GIS entitlement is 0, and the partner's Alw is not 0. In such cases, although the sum of individual benefits calculations is greater than the couple-based benefits entitlement sum, it is considered as an invalid case, as the GIS entitlement is actually invalid for processing the parter's ALW benefit. Therefore, we should only show the couple benefits estimate result on GIS and ALW cards.


MarcoGoC (approved):
approved without testing

PR #557, opened 2023-02-20
Moved fieldset, added legend to fieldset, moved label inside legend

Maxim-Lam:
### Description

Added legend, moved label into legend.

Tell me what you guys think

List of proposed changes:

- moved fieldset
- added legend
- moved label into legend

- inspect form elements with radio button group

### DIFF
--- a/components/Forms/Radio.tsx
+++ b/components/Forms/Radio.tsx
@@ -36,14 +36,16 @@ export const Radio: React.VFC<InputProps> = ({
   return (
     <div className="radio">
       <fieldset>
-        <QuestionLabel
-          name={name}
-          type="radio"
-          label={label}
-          requiredText={requiredText}
-          helpText={helpText}
-        />
-        <Tooltip field={name} />
+        <legend>
+          <QuestionLabel
+            name={name}
+            type="radio"
+            label={label}
+            requiredText={requiredText}
+            helpText={helpText}
+          />
+          <Tooltip field={name} />
+        </legend>
         <div role="radiogroup" className="mt-2.5">
           {values.map((val, index) => (
             <div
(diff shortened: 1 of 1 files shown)


alex-solo:
Were you able to check this with a screen reader? If it's an improvement, then we should go for it. The way we build up the form it's kind of tricky to do, but ideally we would have the fieldset element around the last 2 questions in Age since they both relate to OAS deferral age.

Again, if this is better for the reader let's go with it, let's prioritize just getting it done and passing the ITAO stuff and we can work out the details after.

components/Forms/Radio.tsx
```diff
@@ -36,14 +36,16 @@ export const Radio: React.VFC<InputProps> = ({
...
+        <legend>
+          <QuestionLabel
+            name={name}
+            type="radio"
+            label={label}
+            requiredText={requiredText}
+            helpText={helpText}
+          />
+          <Tooltip field={name} />
+        </legend>
```
MarcoGoC, line 48: Not sure if legend should wrap the tooltip too, as Alex mentions if it works on a screen reader all good

Maxim-Lam: yeah I wasn't sure if it should be there too... It was after question label so I just added there. @alex-solo what do you think?

Maxim-Lam:
> 

totally agree; the way these forms are built are pretty unique. If you guys agree, we can send it like this since the issue raised was lack of legend element for a fieldset and deal with an issue (if it even gets raised) later. @alex-solo  @MarcoGoC  @alexunique0519

alex-solo (approved):
Approving based on discussion

components/Forms/Radio.tsx
Maxim-Lam: resolving convo

PR #562, opened 2023-02-21
ADO-107285: Remove test banner and add info message on the overview page

alex-solo:
### Description

Once again removing the test banner. Instead adding an info message as per ADO-107285

See ADO-107285 to compare against


MarcoGoC (approved):
I do not know about this wasn't this a request to have it there ?

PR #564, opened 2023-02-21
Fixed issues in new OAS results card

alexunique0519:
### Description
This PR addresses multiple fixes related to the OAS result card, also it covers changes in OAS eligibility determination process and clawback calculation, as the old way was to use relevant income (applicant's and partner's). The proper way is to only use applicant's income (client's).  

List of proposed changes:
- ADO-107515
- ADO-107517
- ADO-107518
- ADO-107524


MarcoGoC (approved):
approved without testing

PR #572, opened 2023-02-22
Fix for results

alexunique0519:
Multiple small fixes for both oas and gis results

### DIFF
--- a/i18n/api/fr.ts
+++ b/i18n/api/fr.ts
@@ -363,6 +363,8 @@ const fr: Translations = {
         'Vous recevrez automatiquement un paiement si votre revenu est inférieur à ',
       youShouldReceiveLetter:
         "Vous devriez recevoir une lettre au sujet de votre statut d'inscription le mois après votre 64e anniversaire.",
+      youShouldHaveReceivedLetter:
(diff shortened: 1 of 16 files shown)


MarcoGoC (approved):
Looks good to me, with the question about revenu versus revenus is the first also used for single income?

i18n/api/fr.ts
```diff
@@ -376,7 +376,7 @@ const fr: Translations = {
     },
     gis: {
       eligibleDependingOnIncomeNoEntitlement:
-        'Vous êtes probablement admissible à cette prestation si {INCOME_SINGLE_OR_COMBINED} est moins que {INCOME_LESS_THAN}. Fournissez {YOUR_OR_COMPLETE} revenus pour obtenir une estimation de paiement mensuel.',
+        'Vous êtes probablement admissible à cette prestation si {INCOME_SINGLE_OR_COMBINED} est moins que {INCOME_LESS_THAN}. Fournissez {YOUR_OR_COMPLETE} revenu pour obtenir une estimation de paiement mensuel.',
```
MarcoGoC, line 379: votre revenus / vos revenu 
revenu should be added as part of the variable {your_or_complete}

PR #581, opened 2023-02-23
107942 mixed results

Maxim-Lam:
### Description

fixed app would display the not eligible banner when no income was provided when it should display it.

List of proposed changes:

- Implemented NaN fix
- fix eligibility displayed when income is not provided (see task 107942) 
- 

fill out the form with valid data but do not provide income. 
it should now say you are eligible and display the proper monthly amount

### DIFF
--- a/utils/api/benefits/oasBenefit.ts
+++ b/utils/api/benefits/oasBenefit.ts
@@ -252,8 +252,8 @@ export class OasBenefit extends BaseBenefit<EntitlementResultOas> {
    */
   private get clawbackAmount(): number {
     if (
-      this.input.income.client < legalValues.oas.clawbackIncomeLimit ||
-      !this.input.income.client
+      !this.input.income.client ||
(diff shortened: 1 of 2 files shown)


utils/api/benefits/oasBenefit.ts
```diff
@@ -251,6 +251,7 @@ export class OasBenefit extends BaseBenefit<EntitlementResultOas> {
    * The yearly amount of "clawback" aka "repayment tax" the client will have to repay.
    */
   private get clawbackAmount(): number {
+    if (!this.input.income.client) return 0
```
alexunique0519: can we merge this condition with the following one using ||

alexunique0519 (changes requested):

utils/api/benefits/oasBenefit.ts
Maxim-Lam: good point Yan, checking in the change

alexunique0519 (approved):
LGTM

PR #582, opened 2023-02-23
107923 - Learn more about tax - contradictory changes 

MarcoGoC:
### Description
- there are two contradictory tasks that change the text for the Recovery Tax link. so a new Link variable needed to be created 

List of proposed changes:
- as above. 

- Verify on the [dynamic url](https://ep-be-dyna-107923-clawback.bdm-dev-rhp.dts-stn.com/)


alexunique0519 (approved):
LGTM

PR #583, opened 2023-02-23
Ado 101269

alexunique0519:
### Description

Fixed ado-101269 per discussion with Vero and Lorelei. Also, change the way to generate the first result link in result page.


MarcoGoC (approved):
I think changes here have nothing to do with the original request, but well

PR #584, opened 2023-02-23
Update DS, add CTA, update FormDatePicker

alex-solo:
### Description

This captures ADO-100763 as well as an old one ADO-87399. 

List of proposed changes:

- updated DS
- deleted the tailwind/forms plugin that was taking CSS precedence over DS CSS and breaking FormDatePicker
- had to add CSS to global.css to properly style the CTA component

### Additional Notes

I had a suspicion for a while that something was off with the way we were using global.css to overwrite the styles that were breaking in DS components when they were brought in. In other words, we would import a component, it would not look like promised in DS, so local CSS was used to force it to look like it did in DS storybook. This was not good and of course defeats the purpose of having a design system. 

After investigating and learning more about tailwind, the fix for why the FormDatePicker component was breaking became clear - a form plugin with its opinionated CSS reset was taking precedence (overwriting) the CSS of the component. Since we were not heavily relying on this form plugin, I removed it, adding CSS locally to fix any breaking changes.

I learned also that the "base" directive that we import in global.css is taking precedence over the imported components from DS. "base" or as tailwind calls it, "preflight" CSS (https://tailwindcss.com/docs/preflight) is simply a large CSS reset. However, instead of this order: preflight -> other tailwind -> global.css -> design system
the preflight and other tailwind CSS took precedence over the design system. I have confirmed this with a clean project (next-template) and unfortunately, with a standard tailwind CSS setup, importing design system components doesn't work as expected out of the box - the preflight CSS will overwrite anything that happens to overlap.

 Zain will post a possible solution to this in the design-system channel on Slack in the near future. In order to make sure that CSS for DS components takes precedence, DS will come with its own compiled CSS file that we will import like this (notice, after our global.css) :
Screenshot 2023-02-23 at 1 32 18 PM

For now, to get this working and "out the door" we'll use the latest version of DS and patch breaking CSS as before. Once the release is behind us we will implement these changes that will be present in the DS going forward.


MarcoGoC (approved):
With the exception of the year field allowing symbols, and the month having a grey background on Firefox.  
It fixes the header issue and the footer issues #107778 . 
works fine otherwise

alexunique0519 (approved):
LGTM

PR #585, opened 2023-02-23
ADO-104733

alexunique0519:
### Description

Made change to the partner question section per ADO 104733

List of proposed changes:

- 1. when a partner is younger than 60, the partner will not be eligible for OAS, GIS, ALW
        therefore, no following question will show up except income
        
- 2. when a partner is older or equals to 60, but younger than 65, the partner will not be eligible for 
        OAS, therefore, hides the partner benefit section
        
- 3. when a partner is older or equals to 65, we show the partner benefit section


utils/api/benefitHandler.ts
```diff
@@ -210,24 +210,35 @@ export class BenefitHandler {
...
+        3. when partner is older or equals to 65, we show the partner benefit section
+       */
 
       if (
-        (this.input.client.invSeparated && this.input.partner.age >= 60) ||
-        (this.input.client.partnerBenefitStatus.helpMe &&
-          this.input.partner.age >= 60)
+        this.input.partner.age < 60 ||
+        this.input.partner.age > 123 ||
+        this.input.partner.age === undefined
```
MarcoGoC, line 232: !this.input.partner.age  instead?

alex-solo (approved):
works well now

PR #602, opened 2023-02-27
Fix income

alexunique0519:
### Description

In OAS, previously it was using this.input.income.relevant to compare income, however, in OAS eligibility and entitlement part, we only care about each single person's income. Therefore, this.input.income.relevant had been changed to this.input.income.client. However, the income value was fixed for each person, for example, the client's income will always be the applicant's income, and the partner's income will always be the same no matter how we pass the input variable to the OAS benefit constructor. This change is to make sure we use the correct income value to make the calculation. For example, if the _partner_ variable has been passed as true, it means we only want to calculate the partner's entitlement, not the applicant's.


utils/api/benefits/oasBenefit.ts
```diff
@@ -31,8 +31,12 @@ export class OasBenefit extends BaseBenefit<EntitlementResultOas> {
     // helpers
     const meetsReqAge = this.input.age >= 65
 
+    let income = this.partner
```
alex-solo: Income variable can be `const` here, no? Also, this is used in the clawback calculation below - can extract to constructor and have the income variable on the class itself so can be reused everywhere. Just general comments/suggestions but I will leave it up to you and approve in the mean time

alex-solo (approved):
See code comment

utils/api/benefits/oasBenefit.ts
alexunique0519: Changed income to be one member of OASBenefit class @alex-solo

PR #606, opened 2023-02-28
check age before returning AFS

alexunique0519:
### Description

Add age check before returning gisSituation, the calculation error, in this case, was caused by the wrong gisSituation value, as the applicant is ineligible for AFS, however it still returned AFS. 

- Add age check before returning gisSituation
- remove pre-release-ds, since we don't need it. It added a redundant design system library in the yarn cache.


MarcoGoC (approved):
GIS should not applied if widowed if he is too young

PR #607, opened 2023-03-01
110340 - GIS changes 

MarcoGoC:
### Description
- The combined income could be over the limit, but still is eligible?? makes no sense but I am not making the rules. This changes conflicts with 109965 which basically is any single income is too high return 'income too high' message so now I added the amount to the condition.   

- as above


alexunique0519 (approved):
Tested and works for me, although how to identify if the income is too high is tricky to me.

PR #608, opened 2023-03-02
99999: fixed the exception when income is not provided

alexunique0519:
### Description

This issue is caused by invoking the _const amount = this.formulaResult()_ function before the main check, since the eligibility result doesn't contain any value yet, it will directly call the getEntitlement function,  although the income is not provided.  Please test with multiple scenarios to make sure this fix covers all possibilities.


MarcoGoC (approved):
good catch

PR #609, opened 2023-03-02
account for the current month

alex-solo:
## [NNN](https://dev.azure.com/VP-BD/DECD/_workitems/edit/NNN) (ADO label)
ADO 110737
### Description

- Need to set max age based on the current month

- extract logic of calculating maximum age

ex. Current possible maximum age is 123.2 (January 1900) but only accounting for the birth year the maximum allowed age is 123.

### DIFF
--- a/utils/api/definitions/schemas.ts
+++ b/utils/api/definitions/schemas.ts
@@ -23,6 +23,12 @@ import {
  *
  */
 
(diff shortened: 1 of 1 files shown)


MarcoGoC (approved):
Tested +1

PR #610, opened 2023-03-03
Ado 107432

alexunique0519:
### Description

Fixed the issue ado-107432 by checking if the result is valid, since when _partnerGis.entitlement_ equals 0, which means the client could not get ALW. Therefore, although maybe the amount of the total benefits, in this case, is higher, we should not consider it as a valid case. Instead, we need to choose the opposite option.

- Fixed the case when applicant is eligible for ALW
- Fixed a small issue gis card details showing partner has a '0.00' entitlement

### DIFF
--- a/utils/api/benefitHandler.ts
+++ b/utils/api/benefitHandler.ts
@@ -566,11 +566,20 @@ export class BenefitHandler {
                 .calculatedBasedOnIndividualIncome
             )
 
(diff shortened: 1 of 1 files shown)


MarcoGoC (approved):
I cannot reproduce in current dev version, Lorelei approved the changes

PR #611, opened 2023-03-03
Ado-110925

alexunique0519:
### Description

- Adding the commented part back in fieldClasses.ts, because it matters for the entitlement calculation.


MarcoGoC (approved):
This fixes the issue +1

PR #612, opened 2023-03-03
added test banner to all three pages and changed text

alex-solo:
### Description

- Add "Work in progress" message to all 3 pages and change the text according to ADO 110717

-
-

Make sure the message is on all 3 pages and has correct text in both English and French


MarcoGoC (approved):
It could be added to the Layout/index.ts instead ??

alex-solo:
> 

You are absolutely right, made the change.

PR #616, opened 2023-03-07
Ado 111039

alexunique0519:
### Description
- The reason caused this issue was when the first time we got the gis entitlement, the client.partnerBenefitStatus was only set to help, and after that we get the partner's ALW eligibility. Therefore, the client's gis was not calculated using table-4, to fix this after we get the partner's ALW eligibility, if it is eligible we need to re-calculate the client's GIS.  

- Fixed 111039
- Added setValueForAllResults to clear redundant code.


package.json
```diff
@@ -22,6 +22,7 @@
   "dependencies": {
     "@dts-stn/service-canada-design-system": "1.56.6-dev.bdf8433bd",
     "@tailwindcss/forms": "^0.5.0",
+    "ansi-styles": "^6.2.1",
```
MarcoGoC: ???

alexunique0519: forgot to delete after a workaround

alex-solo (approved):
Looks good, just one merge conflict on the .yarnrc.yml file since it was added in my last PR.

PR #618, opened 2023-03-07
110998 - French text changes

MarcoGoC:
### Description
- some label changes 

- as above


alex-solo (approved):
works as expected

PR #619, opened 2023-03-08
107429 - New case ALW

MarcoGoC:
### Description
- Removing message when partner entitlement result = 0 

- as above


alexunique0519 (approved):
Looks good

PR #620, opened 2023-03-08
Added open a new tab icon to all links

alexunique0519:
### Description

- Added "open a new tab" icon to all links in the app.


styles/globals.css
```diff
@@ -234,4 +237,19 @@
     list-style: decimal;
     list-style-position: inside;
   }
+
+  #oasDefer a,
+  #oasLink0,
```
MarcoGoC, line 242: would it work with just a.generatedLink ?, I think so

MarcoGoC (approved):
just a  comment

styles/globals.css
alexunique0519: The links in the result card don't have this class

PR #621, opened 2023-03-09
Modified logic to display elements only if text is available, not entire object.

Maxim-Lam:
### Description

benefit card was displaying empty paragraph tags when no next step available because it was checking if the object being passed existed. The problem is the object was always passed anso it would just diplay empty p tags.

So instead of checking if the object is there, I changed it to check for the text being there.


MarcoGoC (approved):
+1

PR #622, opened 2023-03-09
111111 - new text

MarcoGoC:
### Description
- removed some text eng and fre pages

- See task for changes


alexunique0519 (dismissed):
Should sync with dev, as the "b" has been replaced with "strong", otherwise looks good to me

alexunique0519 (approved):

PR #624, opened 2023-03-09
Ado 107253

alexunique0519:
### Description

From the ITAO audit report, this issue is described as the screenreader can't tell the tooltip's status, i.e. expand/collapse, when the user clicks the "more information" link. It also suggested using details and summary to implement the tooltip component. Therefore, I have changed the tooltip component according to the recommended method, the basic functionalities are still the same, meanwhile, it also enables the screenreader to tell users the tooltip's current status.

Available on the [dynamic URL](https://ep-be-dyna-ado-107253.bdm-dev-rhp.dts-stn.com/)

### DIFF
--- a/components/Tooltip/tooltip.tsx
+++ b/components/Tooltip/tooltip.tsx
@@ -41,7 +41,7 @@ export const Tooltip: React.FC<{
   if (!tooltipData) return <></>
 
   return (
-    <div
+    /*    <div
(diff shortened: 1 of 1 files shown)


alex-solo (approved):
nice, good call on "details" and "summary" tags. Seems to work as expected and semi-tested by Justin sounds like.

PR #625, opened 2023-03-09
111539 - Reset partnerAge to default to erase previously entered values

MarcoGoC:
### Description
- To reproduce: Select Married or common-law, fill out year of birth. Select Widowed. Select Married or common-law.
The year is still filled out, and has an error 

- Reset date when field is partnerAge AND error message is undefined this avoid reset when coming back from the results page to change the age  

### Additional Notes
- This solution would probably wont work after a fix for 'validate only on button click'


alex-solo (approved):
nice solution, works as expected

PR #626, opened 2023-03-09
ADO-107379: Add decimal to income, adjust width of input

alex-solo:
### Description

- See explanation and mockup in ADO task

- change width of input
- utilize the "isNumericString" prop of react-number-format and use regex selectively to allow decimal separator to be a comma in French

- Should work as expected in terms of calculations
- Visually should look like:

EN: $1,123,432.34
FR: 1 123 432,34 $

### Additional Notes
N/A

### DIFF
--- a/client-state/InputHelper.ts
+++ b/client-state/InputHelper.ts
@@ -36,7 +36,7 @@ export class InputHelper {
 
   setInputByKey(key: FieldKey, newValue: string): void {
     if (newValue === '' || newValue === undefined) delete this.inputs[key]
-    else this.inputs[key] = InputHelper.sanitizeValue(newValue)
+    else this.inputs[key] = InputHelper.sanitizeValue(newValue, this.language)
     this.setInputs(this.inputs)
   }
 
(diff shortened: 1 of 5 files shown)


MarcoGoC (commented):
copy/paste does not work when in wrong format to the locale, 
also entering french format on english page does not work

alex-solo:
> copy/paste does not work when in wrong format to the locale, also entering french format on english page does not work

Can you give a practical example that would cause a problem? 

If you want to check out the demos from `react-number-format`, our input behaves exactly the same: https://s-yadav.github.io/react-number-format/docs/numeric_format

Pasting 123456789 gives us "$123,456,789" in English and "123 456 789 $" in French. Do you think copying English formatted string and pasting to French is something we should worry about? 

Try the demos from `react-number-format`, it's pretty easy to "trick it" with weird pasting. Happy to discuss

alex-solo:
Made some changes to ensure that we always get cents to 2 decimal places but only if user includes decimal. If user enters "234.3", it will change to "234.30" both in the input (on blur) and in the results page.

alexunique0519 (approved):
Tested in both Fr and EN, the formats are correct. Switching between EN/FR, the number in the currency input keeps the value and only changes to the format under that locale, works as expected.

PR #630, opened 2023-03-13
replaced all remaining <b> with <strong>

Maxim-Lam:
replaced all remaining <b> tags with <strong> tags for accessibility purposes.


alexunique0519 (approved):
LGTM

PR #631, opened 2023-03-13
Fixed the   grey background issue in firefox

alexunique0519:
Explicitly define the background color for select elements in the firefox browser, since Firefox gives select controls grey background color by default.


MarcoGoC (approved):
it works on FF

PR #632, opened 2023-03-13
111539 year

MarcoGoC:
### Description
- `error` is a string that returns false when empty the extra condition added just checks for it. 

- as above
-


alexunique0519 (approved):
Tested and year was cleared when switching between married and widowed.

PR #633, opened 2023-03-14
Fixed 107262 following the ITAO suggestion

alexunique0519:
### Description

The original issue was about the screen reader could not provide clear information when users navigate to an input field, the issue is resolved by making changes in QuestionLabel.tsx and NumberField.tsx following the suggestions in ITAO audit report.  I have tested the behavior with both VoiceOver and NVDA, and it works as expected. 

- removed the aria-label attribute in the label element in the QuestionLabel component, so that the label text(i.e. the questions) can be read properly.  
- added _aria-describedby_ to the NumberField component, which enables the screen reader to recognize the help text when navigating to the input field. 

Use screen readers like VoiceOver or NVDA

### DIFF
--- a/components/Forms/NumberField.tsx
+++ b/components/Forms/NumberField.tsx
@@ -65,6 +65,7 @@ export const NumberField: React.VFC<NumberFieldProps> = ({
         enterKeyHint="done"
         allowNegative={false}
         decimalSeparator={null}
+        aria-describedby={`help-text-enter-${name}`}
       />
 
       {error && (
--- a/components/Forms/QuestionLabel.tsx
+++ b/components/Forms/QuestionLabel.tsx
@@ -21,7 +21,6 @@ export const QuestionLabel: React.FC<QuestionLabelProps> = ({
     <>
       <label
         htmlFor={fieldId}
-        aria-label={name}
         data-testid={`${type}-label`}
         className="text-content font-bold inline mb-2.5 mr-2"
       >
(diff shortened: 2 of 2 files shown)


components/Forms/QuestionLabel.tsx
```diff
@@ -33,6 +32,7 @@ export const QuestionLabel: React.FC<QuestionLabelProps> = ({
       </label>
       {helpText && (
         <span
+          id={`help-text-${fieldId}`}
```
MarcoGoC: un-intented consequences? 
image

components/Forms/NumberField.tsx
```diff
@@ -65,6 +65,7 @@ export const NumberField: React.VFC<NumberFieldProps> = ({
         enterKeyHint="done"
         allowNegative={false}
         decimalSeparator={null}
+        aria-describedby={`help-text-enter-${name}`}
```
MarcoGoC, line 68: I can't find and id = `help-text-enter-xxx`. maybe you meant `helpText-xxx` ?

MarcoGoC (changes requested):
see comments

components/Forms/NumberField.tsx
alexunique0519: Screenshot 2023-03-14 at 1 32 51 PM

alexunique0519: @MarcoGoC

components/Forms/QuestionLabel.tsx
alexunique0519: Fixed by checking the fieldId value

MarcoGoC (approved):
wave shows some orphan labels not sure how would that works on a screen reader

PR #638, opened 2023-03-16
Add new icon

alexunique0519:
### Description

- As per ado-87355, the "open a new tab" has been updated, therefore replaced the old one with a new one in our app. 

- change the icon (it looks ugly tho)
- replace the LINK component from DS to anchor element, because the abbr element inside the Link component broke the style.


MarcoGoC (commented):
Looks good to me, however it needs to be merged with develop first

MarcoGoC (approved):
Thanks

PR #642, opened 2023-03-17
Add sanity tests

alexunique0519:
### Description

Added all sanity test cases into a separate test file - sanity.test.ts. Every time we commit a change, the sanity test cases will be executed, and this is to ensure in the future we don't bring breaking changes to the app. 

- created sanity.test.ts
- added legalValues_test json file, and this file is for testing only, it will not be updated in future 
- made some code changes for expectUtils, so it can be used to verify partner's benefit results.


MarcoGoC (approved):
this looks good to me

PR #646, opened 2023-03-21
Changed number formatter to not display .00 but keep the 2 decimal points for only 1 decimal number present

Maxim-Lam:

- no longer displays .00 
- decimal points such as .1 or .5 etc are still displayed as .10 and .50

Enter income as xx.00

notice income on next page does not display .00

enter income as xx.x

notice income still display as xx.x0


alex-solo (approved):
Nice!

PR #647, opened 2023-03-22
Made the age fields shorter in length

Maxim-Lam:
### Description

Made the age field smaller

Made the age field smaller

Check the age field, notice that is smaller


alexunique0519 (approved):
LGTM

PR #648, opened 2023-03-22
[Until OK is giving to merge] 113491 - Change to Resident question and tests

MarcoGoC:
### Description
- Change the Residence question

- Since the name of the variable does not make sense with the new values it was changed from LivedOutsideCanada to LiveOnlyInCanada
- Since the question changed from Negative to a Positive all the test were change to the opposite `true to false` and `false to true`

- Better to leave it to the testers as there were lots of changes


alex-solo (approved):
looks good

utils/api/benefitHandler.ts
```diff
@@ -187,9 +187,9 @@ export class BenefitHandler {
       FieldKey.LIVING_COUNTRY,
       FieldKey.LEGAL_STATUS,
       FieldKey.MARITAL_STATUS,
-      FieldKey.LIVED_OUTSIDE_CANADA,
+      FieldKey.LIVED_ONLY_IN_CANADA,
     ]
-    if (this.input.client.livedOutsideCanada) {
+    if (!this.input.client.livedOnlyInCanada) {
```
alexunique0519: Some changes need to be made here, so for now the year input shows when the card is open. It should only show when a user selects 'No' for 'Since the age of 18, have you only lived in Canada?'

alexunique0519 (dismissed):

utils/api/benefitHandler.ts
MarcoGoC: fixed

alexunique0519 (approved):
LGTM

PR #654, opened 2023-03-24
Auth added as requested by Amy

MarcoGoC:
### Description
- Followed instructions from Eric to add an ssh password


MarcoGoC:
these changes don't work as these are for nginx which the "hardware" isn't using

PR #660, opened 2023-03-28
Added next-auth to add authentication to all pages

alex-solo:
### Description

- Added next-auth  with basic credentials (username and password discussed in latest Teams meeting). Ask in Slack if you need it

- Wrap components in Auth component that checks the signin status based on what the current pathname is
- In case we add new pages or want to make just the home page public, we can do so

log in with provided username and password and the application should behave as normal after that. Check cookies in browser, the tokens should be there

### Additional Notes

I seem to lose the session token sometimes - when I leave the window for an extended period of time. That might be due to my Chrome settings but keep an eye out for that in case it happens too often. Ideally, once you have the token, you shouldn't have to sign back in until its expiry (30 days by default)


MarcoGoC (approved):
works

PR #669, opened 2023-03-30
115152 - adding partner alw amount

MarcoGoC:
### Description
- Partner amount was not being loaded to the card detail 

- If eleigible and amount > 0 add the message for partner alw


alex-solo (approved):
Can remove console.log if we don't need it otherwise, looks good!

PR #670, opened 2023-03-31
115291 - Links

MarcoGoC:
### Description
- 1 seems the bookmarks aren't working because of the Auth component
- 2 for the links removing underline from the class works for some reason.   

- as above


Maxim-Lam (approved):
Looks good

PR #672, opened 2023-04-03
Only request username and password when deploying to dev and staging

alex-solo:
### Description

- We want to protect only the dev and staging URLs for now. Everything should behave the same way but will be different (no auth) when on SC Labs.

### Additional Notes

TODO: When I gain access to prod TeamCity, will add "production" as the app variable.


MarcoGoC (approved):
lgtm

PR #673, opened 2023-04-04
Use nextjs link component to avoid reloading the page

alex-solo:
### Description

- Unfortunately DSLink doesn't work the way we want under the hood - it uses an "a" tag which reloads the page. I did not notice this until now. We want to delegate switching pages to JavaScript in true Single Page Application fashion. 

- Use NextJs Link component to navigate back to eligibility page from the results page

Click on edit link on results page and ensure the link goes back to the correct location on the eligibility page.

### Additional Notes

TODO: change all links to NextJS Link components AND/OR advise DS team to utilize Link component in their project.

### DIFF
--- a/components/ResultsPage/YourAnswers.tsx
+++ b/components/ResultsPage/YourAnswers.tsx
@@ -1,10 +1,10 @@
-import { Link as DSLink } from '@dts-stn/service-canada-design-system'
 import { FieldInput } from '../../client-state/InputHelper'
 import { numberToStringCurrency } from '../../i18n/api'
 import { WebTranslations } from '../../i18n/web'
 import { BenefitHandler } from '../../utils/api/benefitHandler'
 import { FieldConfig, FieldType } from '../../utils/api/definitions/fields'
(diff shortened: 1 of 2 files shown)


components/ResultsPage/YourAnswers.tsx
```diff
@@ -58,13 +58,14 @@ export const YourAnswers: React.VFC<{
...
                 </div>
                 <div className="justify-self-end self-end">
-                  <DSLink
-                    id={`edit-${input.key}`}
-                    href={`eligibility#${input.key}`}
-                    text={tsln.resultsPage.edit}
-                    target="_self"
-                    ariaLabel={tsln.resultsEditAriaLabels[input.key]}
-                  />
+                  <Link href={`eligibility#${input.key}`}>
```
Maxim-Lam, line 61: Wrap an a tag with a Link component?

Maxim-Lam (changes requested):

Maxim-Lam (approved):
Looks Good!

PR #675, opened 2023-04-05
115850 - GIS update

MarcoGoC:
### Description
- Message was being loaded clientGis and later in the code being overwritten at line 980 

- by loading it to allResults it avoids being overwritten


alex-solo (approved):
GIS calculated using single income message shows up as expected

PR #677, opened 2023-04-05
ADO 116145: Show correct message when single income is above threshold

alex-solo:
### Description

- When only one income is provided we need a different message if that one income is already above the threshold

Unrelated changes:
- added more env vars to the sample env file we have
- deleted unused import

Refer to the task in ADO for expected behaviour


MarcoGoC (approved):
Good +1

PR #683, opened 2023-04-11
ADO Limit length of input

alex-solo:
### Description

- Limit the length of input so we don't have the "unsafe number" error.

-
-


MarcoGoC (approved):
maybe limit max of numberField to 3??? years living in canada can't  be millions of years

PR #685, opened 2023-04-12
Fix inv sep, when switching to widowed/single

alex-solo:
### Description

-

-
-


MarcoGoC (approved):
Good find thanks

PR #687, opened 2023-04-13
117792 - ALW change message when married

MarcoGoC:
### Description
- Display different message when single/widow than when married 

- as above


Maxim-Lam (approved):
LGTM

PR #688, opened 2023-04-13
Partner eligible was being added in two different locations for ALW. Removed it from BenefitHandler left it in alwBenefit

Maxim-Lam:
### Description

- parnter eligibility for ALW was being added twice in two different locations.

- remove partner benefit eligibility from being added twice to ALW benefit card.
- it was being added in alwBenefit.ts and benefitHandler.ts. Removed from benefitHandler.ts

enter information on bug 117783. Scroll down to Allowance section on result page. no more duplicate


MarcoGoC (approved):
looks good

PR #690, opened 2023-04-18
Fix for GIS message when inv seperated

Maxim-Lam:
### Description

- fix messagy for when eligible for GIS but inv seperated

- same as description

I will post a screenshot with parameters to enter to get the result

### Additional Notes

open to feedback if there is a better way

### DIFF
--- a/utils/api/benefitHandler.ts
+++ b/utils/api/benefitHandler.ts
@@ -938,6 +938,17 @@ export class BenefitHandler {
             allResults.client.gis.entitlement.result = applicantGisResultT1
             allResults.client.gis.entitlement.type = EntitlementResultType.FULL
           }
+          if (
+            this.input.partner.invSeparated &&
+            clientGis.entitlement.result > 0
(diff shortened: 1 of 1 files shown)


Maxim-Lam:
image

utils/api/benefitHandler.ts
```diff
@@ -991,6 +991,17 @@ export class BenefitHandler {
             '--- both are not eligible for alw - applicant oas = 0 & partner oas > 0 --- end'
           )
         }
+        if (
```
MarcoGoC: good this works, but move the code to before **line 940**, if you look at the console messages you'll see why.

MarcoGoC (commented):
see comment

utils/api/benefitHandler.ts
MarcoGoC: edit: line 940.

MarcoGoC (approved):
thanks

PR #695, opened 2023-04-25
Added cypress to app

alex-solo:
### Description

- Added cypress for end to end testing. 
- Used 12.9.0 instead of latest 12.10.0 due to a dependency error it was throwing
- Making cypress work with Yarn 3 (which uses PnP by default) has been discussed a lot on GH and the team has been working on making cypress compatible up until recently. The latest discussion is just weeks old. 
- Using cypress.config.ts only works if using node_modules so I switched to JS here.

- Try script commands from package.json that relate to cypress. Cypress runner should launch, choose end to end testing and pick browser.
- The current tests will fail because we have a login screen. 
- Next task is to write actual tests

### Additional Notes

You will probably see an error when launching cypress:
"Most NODE_OPTIONs are not supported in packaged apps. See documentation for more details."

This is due to Yarn PnP but seems to not affect function.


Maxim-Lam (approved):
Nice job alex, works great

PR #698, opened 2023-05-01
ADO-116435. Make the max length limit the same for EN and FR

alex-solo:
### Description

- Because EN and FR are displayed differently (leading $, spaces, commas, etc.) the same max length doesn't show the same amount in FR and EN. Now this is fixed. Both EN and FR currency fields are limited to 11 NUMBERS.


MarcoGoC (approved):
+1

PR #711, opened 2023-05-24
123563 - fix to duplicate cardCollapse for GIS

MarcoGoC:
### Description
- adding Partner to if to avoid duplicate message

- as above
-


Maxim-Lam (approved):
LG

PR #712, opened 2023-05-24
add meta tags to index page

alex-solo:
### Description

- Added common metatags to the Head component in Layout
- Tags that are unique to the questions and result pages are added to those pages only

- check in page source

### Additional Notes
- There is a requirement to make the "modified date" dynamic which will require adding an environment variable in TeamCity. After this, we can read the env var in the app. Another task will be made for that

### DIFF
--- a/components/Layout/Head.tsx
+++ b/components/Layout/Head.tsx
@@ -1,3 +1,4 @@
+import { is } from 'cypress/types/bluebird'
 import NextHead from 'next/head'
 import { useRouter } from 'next/router'
 import { WebTranslations } from '../../i18n/web'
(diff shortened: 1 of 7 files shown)


components/Layout/Head.tsx
```diff
@@ -1,13 +1,73 @@
...
+            : `http://ep-be.alpha.servicecanada.ca/${tsln._language}${router.pathname}`
+        }
+      />
+      <link
+        rel="alternate"
+        lang="en"
+        href={
+          isCanadaDotCa
+            ? `oas-estimator.service.canada.ca/${tsln._language}${router.pathname}`
+            : `shttps://ep-be.alpha.service.canada.ca/${tsln._language}${router.pathname}`
```
MarcoGoC: typo?   `shttps:`

alex-solo: Good catch, yes!

MarcoGoC (approved):

PR #714, opened 2023-05-25
ADO-123324 - Fix dynamic options population

alex-solo:
### Description

- When a maxYear is reached we dynamically populate months select to limit to the max number of months possible. Otherwise the default limit for months is 11.

So if age is 67.5, we can select 1 year and 11 months deferral. When selecting "2" for years, the months select is limited to 6 months only

Test changing age, changing year and month options


Maxim-Lam (approved):
Nice, works great!

PR #715, opened 2023-05-26
123020 - SuperGIS

MarcoGoC:
### Description
- if receiving partial OAS (15 yrs in canada) and older than 75 yrs old
- receive OAS amount + 10% in GIS 

- added calculation to entitlement formula
- Several Unrelated fixes for situations that had not been tested before 
- Lorelei already check the tests she requested, there may be more than have not been tested


Maxim-Lam (approved):
Tested, good

PR #716, opened 2023-05-26
Fix alternate meta tags

alex-solo:
Quick fix to the "alternate" meta tags. They are supposed to show "fr" when current language is "en" and vice versa


MarcoGoC (approved):
+1

PR #724, opened 2023-05-30
ADO-123324-fix: Max months displayed in select now based on actual birth year and month

alex-solo:
### Description

- Before, I was basing the number of months on the age but because the age is a decimal (1-10), an age of 67.3, represented both months of January and February for example. We needed to compare current month to the month of the client's birth date. 

- When client sets their birth date, a new state for ageDate is passed on to the Duration component
- Because the age property doesn't change when we switch from months January to February (both 67.3 as an example), rerender when the ageDate property changes as well

- When age ends with .8 or .3, make sure the number of months in the "months" select changes when changing birth months. 
- more details and examples in ADO task


MarcoGoC (approved):
checked on the dyna link

PR #730, opened 2023-05-30
Fix Age component to not display when age is out of bounds

alex-solo:
### Description

- Avoid showing negative age and age in the hundreds and thousands

### DIFF
--- a/components/Forms/Age.tsx
+++ b/components/Forms/Age.tsx
@@ -13,7 +13,7 @@ const Age: FC<AgeProps> = ({ age, name }) => {
     if (name === 'age') {
       return tsln._language === 'en' ? 'You are:' : 'Vous avez :'
     } else if (name === 'partnerAge') {
-      return tsln._language === 'en' ? 'Your partner is' : 'Votre conjoint a'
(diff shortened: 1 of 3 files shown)


components/Forms/Age.tsx
```diff
@@ -11,14 +11,14 @@ const Age: FC<AgeProps> = ({ age, name }) => {
   const tsln = useTranslation<WebTranslations>()
   const getLabel = () => {
     if (name === 'age') {
-      return tsln._language === 'en' ? 'You are' : 'Vous avez'
+      return tsln._language === 'en' ? 'You are:' : 'Vous avez :'
```
MarcoGoC, line 14: **You are:** should this be added to the translation files instead?

components/Forms/Age.tsx
```diff
@@ -11,14 +11,14 @@ const Age: FC<AgeProps> = ({ age, name }) => {
...
     } else if (name === 'partnerAge') {
       return tsln._language === 'en' ? 'Your partner is' : 'Votre conjoint a'
```
MarcoGoC: **partner is'**  this might need a colon too ':'

MarcoGoC (approved):

PR #731, opened 2023-05-30
ADO-125539: Calculate oas deferral amounts by months not years

alex-solo:
### Description

- The input needs to save the age as a more precise float, that way we can more accurately calculate the number of months to a given age. We need to do that to calculate the oas deferral amount by the number of months and not years. Before if you were 68, we were using 24 months to calculate the amount at 70 but if the user is actually 68.33, then we can say that they are actually 20 months away and calculate the amount accordingly

- user is 68.33 (Jan 1955)
- 20 months left until they're 70, the amount should be $773.92 and NOT $790.50


MarcoGoC (approved):
checks out

PR #735, opened 2023-06-01
124578 - alw msg

MarcoGoC:
### Description
- Change for another scenario

- as comments on task


alex-solo (approved):
looks good based on the inputs provided in the ADO task and Andreanne seemed to have tested it too and it looks good!

PR #736, opened 2023-06-02
change bold french text, made margins match other headings

Maxim-Lam:
### Description

- update based on feedback in 125588

-
-


alex-solo (approved):
Great!

PR #738, opened 2023-06-02
125655 - changes to force partened to get more money

MarcoGoC:
### Description
- Changes in the black box to force the usage of 'table 3'   
- see the task for the scenario that it fixes
- Since Yan used Table 2, I suspect this change will create more issues but ...   

- as per task 
-

### DIFF
--- a/utils/api/benefitHandler.ts
+++ b/utils/api/benefitHandler.ts
@@ -537,17 +537,38 @@ export class BenefitHandler {
           maritalStatus = new MaritalStatusHelper(MaritalStatus.PARTNERED)
 
           // applicant gis using table2
-          const applicantGisResultT2 = new EntitlementFormula(
+          // const applicantGisResultT2 = new EntitlementFormula(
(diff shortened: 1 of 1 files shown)


alex-solo:
> ## [125655](https://dev.azure.com/VP-BD/DECD/_workitems/edit/125655) Not using Table 3
> ### Description
> * Changes in the black box to force the usage of 'table 3'
> * see the task for the scenario that it fixes
> * Since Yan used Table 2, I suspect this change will create more issues but ...
> 
> #### List of proposed changes:
> * as per task
> 
> ### What to test for/How to test
> ### Additional Notes

Hmm let's be more careful with this one before merging. We want to make sure we're surgically affecting just the scenario from the ADO task without compromising others. If for all the scenarios where Table 2 was being used, we're now using Table 3 that would cause a lot of issues. Let's find out what this would affect before merging.

alex-solo (approved):
Approving based on Lorelei's testing. Nice work!

PR #742, opened 2023-06-06
ADO 123408: Add partner visibility

alex-solo:
### Description

- Add partner visibility as per ADO task.
- If client or partner chooses to not provide income, currently you'll see an empty GIS item but that will soon be impossible when income is mandatory

- See ADO task for how we want the results page to look

Test if both client and partner estimates are appearing as per ADO task. Also note the changes that we want in the links list

### DIFF
--- a/components/ResultsPage/EstimatedTotal.tsx
+++ b/components/ResultsPage/EstimatedTotal.tsx
@@ -28,8 +28,7 @@ export const EstimatedTotal: React.VFC<{
 
   const headerSentence =
     summary.entitlementSum != 0
-      ? tsln.resultsPage.yourEstimatedTotal +
-        numberToStringCurrency(summary.entitlementSum, language)
+      ? tsln.resultsPage.yourEstimatedTotal
       : tsln.resultsPage.yourEstimatedNoIncome
 
   return (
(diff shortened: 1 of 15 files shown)


components/ResultsPage/EstimatedTotal.tsx
```diff
@@ -9,48 +9,63 @@ import { EstimatedTotalItem } from './EstimatedTotalItem'
...
+        return state === SummaryState.AVAILABLE_DEPENDING
+          ? tsln.resultsPage.ifIncomeNotProvided
+          : null
+      }
+
+      if (type === 'header') {
+        return tsln.resultsPage.yourEstimatedTotal
+      }
+    }
+  }
```
MarcoGoC, line 32: switch type
 case 'header'
    return partner ? xxx : yyy

MarcoGoC (approved):
looks good

components/ResultsPage/EstimatedTotal.tsx
alex-solo: good catch! I was going to go back to overwrite it to make it prettier but missed it. Will fix shortly

components/ResultsPage/index.tsx
```diff
@@ -41,8 +50,8 @@ const getEligibilityText = (
   key: string
 ): string => {
   return getEligibility(resultsArray, key)
-    ? `${apiTsln.benefit[key]}: ${apiTsln.result.eligible}`
-    : `${apiTsln.benefit[key]}: ${apiTsln.result.ineligible}`
+    ? `${apiTsln.benefit[key]}`
+    : `${apiTsln.benefit[key]}`
```
MarcoGoC: ? same result either way. you prob meant something different

alex-solo: another good catch! That whole method is actually unneccesary now, so it was deleted

components/ResultsPage/EstimatedTotal.tsx
alex-solo: done

PR #743, opened 2023-06-06
ADO 125332: Make income mandatory for both client and partner

alex-solo:
### Description

- Changes according to ADO task 125332

- Make income mandatory for both client and partner

Please see the task for text and visual changes in both English and French

### DIFF
--- a/i18n/api/fr.ts
+++ b/i18n/api/fr.ts
@@ -141,6 +141,10 @@ const fr: Translations = {
   questionHelp: {
     [FieldKey.INCOME_AVAILABLE]:
       'Fournir votre revenu vous donnera des résultats plus précis.',
+    [FieldKey.INCOME]:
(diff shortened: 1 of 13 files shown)


i18n/api/fr.ts
```diff
@@ -142,10 +141,16 @@ const fr: Translations = {
   questionHelp: {
     [FieldKey.INCOME_AVAILABLE]:
       'Fournir votre revenu vous donnera des résultats plus précis.',
+    [FieldKey.INCOME]:
+      `Incluez le revenu après impôt de toutes les sources (la ligne&nbsp;23600 de votre déclaration de revenus), moins&nbsp;:` +
```
MarcoGoC, line 145: using   ` \`   instead of  `+`   ??

MarcoGoC (approved):
I'll approve it but I disagree with the changes. there is a lot of code that needs to be revised/removed

PR #745, opened 2023-06-07
Small fixes for my previous PRs

alex-solo:
### Description

- Couple of small fixes that were missed from my previous 2 PRs

- styling to include flex so text doesn't wrap to start of line
- french translations that were missed
- OAS in french should have lower case "pension"


MarcoGoC (approved):
seem straight forward

PR #749, opened 2023-06-08
Add AA beacons

alex-solo:
Add AA beacons to more information spans as well as estimate my benefits submit button


Maxim-Lam (approved):
Looks good!

PR #758, opened 2023-06-12
Add font awesome and CTA component

alex-solo:
### Description

- Add the CTA component to our code base to change the icon. For this I had to download the JS packages to be able to use font awesome. 

Check the ADO task to see how the page should look in English and French


package.json
```diff
@@ -26,6 +26,10 @@
   },
   "dependencies": {
     "@dts-stn/service-canada-design-system": "1.57.1-dev.50f5be381",
+    "@fortawesome/fontawesome": "^1.1.8",
```
MarcoGoC, line 29: all of these for an icon change

MarcoGoC (approved):

package.json
alex-solo: Yep, no big deal - they're small js files and the icons are imported one by one, so only one icon has been imported

PR #764, opened 2023-06-15
Fix GIS indv income message not showing

Maxim-Lam:
### Description

- Fix for message not showing up in GIS card

- Check ADO card for params
-

### DIFF
--- a/utils/api/benefitHandler.ts
+++ b/utils/api/benefitHandler.ts
@@ -883,7 +883,6 @@ export class BenefitHandler {
                 this.translations.detailWithHeading.partnerEligible
               )
               if (
-                this.input.partner.invSeparated &&
                 allResults.partner.gis.entitlement.result > 0 &&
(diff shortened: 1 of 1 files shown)


utils/api/benefitHandler.ts
```diff
@@ -882,6 +882,15 @@ export class BenefitHandler {
               allResults.partner.gis.cardDetail.collapsedText.push(
                 this.translations.detailWithHeading.partnerEligible
               )
+              if (
+                this.input.partner.invSeparated &&
```
MarcoGoC: not needed the whole block is for invSeparated = true

alex-solo (approved):
looks good, tested with a couple of scenarios

utils/api/benefitHandler.ts
Maxim-Lam: fixed

PR #766, opened 2023-06-16
[WIP] ADO-128475: Missing ALW estimate for user

alex-solo:
### Description

- Add a new variable to the Alw class that determines the "relevant" income. Relevant income is either combined or individual's income if they're involuntary separated.

-
-


utils/api/benefitHandler.ts
```diff
@@ -900,6 +900,11 @@ export class BenefitHandler {
                   this.translations.detailWithHeading
                     .calculatedBasedOnIndividualIncome
                 )
+
```
MarcoGoC, line 903: good catch 
this works however when combined with the 128459 it display twice the message your partner maybe eligible

MarcoGoC (approved):
tested looks good but requires fixes to 128459

PR #767, opened 2023-06-16
ADO-128459: Missing partner GIS estimate in card

alex-solo:
### Description

- Show partner eligible dropdown whether invSeparated or not
- As long as there is a non-zero estimate, we should show that value in the card

### DIFF
--- a/utils/api/benefits/gisBenefit.ts
+++ b/utils/api/benefits/gisBenefit.ts
@@ -322,11 +322,9 @@ export class GisBenefit extends BaseBenefit<EntitlementResultGeneric> {
             this.translations.detailWithHeading.partnerEligibleButAnsweredNo
           )
         } else {
-          if (!this.input.invSeparated) {
-            cardCollapsedText.push(
-              this.translations.detailWithHeading.partnerEligible
-            )
-          }
+          cardCollapsedText.push(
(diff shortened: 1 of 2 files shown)


utils/api/benefits/gisBenefit.ts
```diff
@@ -322,11 +322,9 @@ export class GisBenefit extends BaseBenefit<EntitlementResultGeneric> {
             this.translations.detailWithHeading.partnerEligibleButAnsweredNo
           )
         } else {
-          if (!this.input.invSeparated) {
```
MarcoGoC, line 325: there is a case for which this was necessary but i cannot find it

MarcoGoC (approved):
with caveats can't find the scenario

utils/api/benefits/gisBenefit.ts
MarcoGoC: found it 
image

MarcoGoC (changes requested):
see scenario

utils/api/benefits/gisBenefit.ts
alex-solo: Good memory, I forgot about that duplicate scenario! Thanks

PR #768, opened 2023-06-16
ADO-128459: Missing partner GIS on card

alex-solo:
### Description

- Delete anywhere in benefitHandler where we add the "partner may be eligible" message and only add it once in gisBenefit for any case when the estimate is > 0 and whether the couple is involuntary separated or not.

### DIFF
--- a/utils/api/benefitHandler.ts
+++ b/utils/api/benefitHandler.ts
@@ -636,19 +636,6 @@ export class BenefitHandler {
               true
             )
 
-            // adds partner calculation when InvSeparated = true
-            if (
-              partnerGis.entitlement?.result > 0 &&
-              (this.input.partner?.partnerBenefitStatus.value ===
-                PartnerBenefitStatus.OAS_GIS ||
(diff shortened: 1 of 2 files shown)


MarcoGoC (approved):
approved with caveats there are some scenarios in which the amount to be used for replacement was different than the calculated. 
can't find the bug  
but maybe the flag useT1versusT3 changed all that and the correct amount is replaced

alex-solo:
> approved with caveats there are some scenarios in which the amount to be used for replacement was different than the calculated. can't find the bug but maybe the flag useT1versusT3 changed all that and the correct amount is replaced

Ok I will hold off to discuss with you in the morning - perhaps you've found a scenario for which it doesn't work? Although I can't picture it. The value for PARTNER_BENEFIT_AMOUNT is replaced with the actual entitlement result. Just where the function is called has changed. (gisBenefit once instead of benefitHandler 3 times). But let me know if you see something I don't

PR #769, opened 2023-06-19
ADO 128475: Missing ALW for clients

alex-solo:
### Description

- When single income is more advantageous, we should use that to calculate ALW 

- more details and scenarios to test against are listed in the ADO task
-

Any additional edge cases you can think of, please go ahead and test

### DIFF
--- a/utils/api/benefitHandler.ts
+++ b/utils/api/benefitHandler.ts
@@ -391,6 +391,7 @@ export class BenefitHandler {
       allResults.partner.gis.entitlement !== undefined &&
       allResults.client.alw.entitlement !== undefined
     ) {
+      console.log('INSIDE OVERWRITE CONDTIONAL')
       if (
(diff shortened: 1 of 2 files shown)


utils/api/benefitHandler.ts
```diff
@@ -891,6 +891,38 @@ export class BenefitHandler {
                     .calculatedBasedOnIndividualIncome
                 )
 
+              // If client is eligible for ALW, need to recalculate estimate based on individual income
+              if (clientAlw.eligibility.result === 'eligible') {
+                const tempClientAlw = new AlwBenefit(
```
MarcoGoC: not sure is necessary, because the correct amount is on applicantAlwCalcSingle

MarcoGoC (approved):
fixes the main 'bug' as requested

PR #770, opened 2023-06-19
128426 - GIS scenario 3 

MarcoGoC:
### Description
- added condition to push message 'based on indiv...'
- Scenario 3 only 

- as above


Maxim-Lam (approved):
perfect, thanks Marco

PR #774, opened 2023-06-21
Partner GIS card missing "partner may be eligible"

alex-solo:
### Description

- Partner may be eligible collapsedText was only being added in the beginning after GIS was calculated but since there are a bunch of overwrites, after this scenario, the GIS for partner changed so we need to add the text to the card. I think this is ok to patch for now, but we should do this only in one place perhaps when a refactoring is under way.


MarcoGoC (approved):
It makes sense but there are some scenarios where it won't work

PR #779, opened 2023-06-21
129559 show partner alw

MarcoGoC:
### Description
- Fix issue discovered where partner alw is not showed on the card
- 
image

- as per images

image


Maxim-Lam (approved):
LG. Maybe remove the console logs if they are not necessary?

PR #781, opened 2023-06-22
 i18n that seems to work

MarcoGoC:
### Description
- i18n that seems to work 

- many besides the actual routing 
- change of the page name to questions as requested by andreanne


alex-solo (dismissed):
When going back  from results page to questions via "Edit your answers", it's not working.

Maxim-Lam (approved):
LGTM. Tested on dyna url and everything seems good

PR #784, opened 2023-06-26
ADO-129794: Updated scraper and added new legal values

alex-solo:
### Description

- Add an option to run the script from JSON data that's stored locally but leave the option of doing so from a URL

-
-


MarcoGoC (approved):
looks good

PR #786, opened 2023-06-27
129799 - i18n - part 2 - fr to use Resultants. 

MarcoGoC:
### Description
- change french page name to resultats using rewrites 

- as above


alex-solo (dismissed):
When on the French results page, toggling back to English, gives a 404

PR #787, opened 2023-06-27
fix for 129703 when editing months

Maxim-Lam:
### Description

- fixed for the OAS defer month dropdown value being updated when changing the year dropdown

- Please see the bug for a better description and steps to reproduce..

- see above


MarcoGoC (approved):
looks good

PR #789, opened 2023-06-28
129804 duration

MarcoGoC:
### Description
- update the state value by calling onChange with the value loaded from the storage

- as above


Maxim-Lam (approved):
Works great!

PR #792, opened 2023-06-29
ADO-129804: Duration component; fix ageDate so the year is not undefined

alex-solo:
### Description

- The problem was in the year being "undefined" when the component re-renders 

During refactor, we should save both "age" as well as month and year input instead of calculating month and year from age. This should do for now.

### DIFF
--- a/components/Forms/Duration.tsx
+++ b/components/Forms/Duration.tsx
@@ -62,7 +62,6 @@ const Duration: FC<DurationProps> = ({
   useEffect(() => {
     if (name in sessionStorage) {
       setDurationInput(JSON.parse(sessionStorage.getItem(name)))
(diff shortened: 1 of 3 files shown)


MarcoGoC (dismissed):
still does not work, issue is not the undefined when rendering but not calling the onChange based on age change or receiving.

alex-solo:
> still does not work, issue is not the undefined when rendering but not calling the onChange based on age change or receiving.

console log the "ageDate" prop - we were getting "undefined" on the birth year. I missed the other scenarios when changing birth month/year so I will address those as well but this addresses the undefined issue

alex-solo (commented):
Has been tested by Justin, it's working now

MarcoGoC (approved):
works now

PR #794, opened 2023-06-29
129200 - alpha sync - updating breadcrumbs for alpha/prod

MarcoGoC:
### Description
- 2 sets of breadcrumbs. one for prod one for alpha 

- as above


Maxim-Lam (approved):
LGTM

PR #797, opened 2023-06-30
i18n - adding default to config

MarcoGoC:
### Description
- default translations to english when locale = default during build 

- I tried a few things as you can see below on each commit but the final fix was to default translations to english


Maxim-Lam (approved):
All good

PR #802, opened 2023-07-06
Updated legal status list  helper

Maxim-Lam:
### Description

- Updated the legal status helper in the legal status section of the form

Change based on ADO activity

- go to form
- go to legal status section
- check that indigenous is now under canadian citizen

### DIFF
--- a/i18n/tooltips/en.ts
+++ b/i18n/tooltips/en.ts
@@ -41,10 +41,10 @@ const en: TooltipTranslations = {
           </p> \
           <ul style='list-style-type: disc; padding-bottom: 8px; padding-left: 20px;'> \
             <li>a Canadian citizen</li> \
+            <li>an Indigenous person registered under the <em>Indian Act</em></li> \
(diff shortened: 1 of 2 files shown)


i18n/tooltips/en.ts
```diff
@@ -41,10 +41,10 @@ const en: TooltipTranslations = {
           </p> \
           <ul style='list-style-type: disc; padding-bottom: 8px; padding-left: 20px;'> \
             <li>a Canadian citizen</li> \
+            <li>an Indigenous person registered under the <em>Indian Act</em></li> \
```
MarcoGoC, line 44: if this was changed for political reasons. indigenous  persons were here first, they should be first

MarcoGoC (approved):

i18n/tooltips/en.ts
Maxim-Lam: Very true lol

PR #804, opened 2023-07-07
129799 - i18n - part 3 - changing afs to alws 

MarcoGoC:
### Description
- Andreanne requested to change afs to alws this has great implications, I am trying to not change every mention of afs just the minimum. 

- as above


alex-solo (dismissed):
This seems to have code from part 2. Maybe combine into one PR so and address the comment that I added on the other one. That was when pulling code, all URLs outlined in task 129799 work. Here, we have the same issue when toggling language on the results page and also can have a scenario where url is "/fr/results#alws" when language is set to French (should be "/fr/resultats#alws")

alex-solo (approved):
looks good. Need to test all ALWS scenarios to make sure nothing was missed but I think yo got it all

PR #810, opened 2023-07-11
131830 - Updated messages for oas

MarcoGoC:
### Description
- different message when estimate = 0

- as per task


Maxim-Lam (approved):
LG

PR #811, opened 2023-07-11
Update messages for GIS and added message for new scenario

Maxim-Lam:
### Description

- updated message for gis card when receiving OAS and added for new scenario

- see description

test for when GIS entitlement > 0 and receivingOas and GIS ent. = 0 and receivingOas for new messages


MarcoGoC (approved):
good

PR #813, opened 2023-07-13
131636 - update license 

MarcoGoC:
### Description
- updating license to GPL3

- as above


Maxim-Lam (approved):
good

PR #826, opened 2023-07-15
ADO-132333, 132303: added two changes based on regression testing

alex-solo:
### Description

-

-
-

### DIFF
--- a/components/ResultsPage/WillBeEligible.tsx
+++ b/components/ResultsPage/WillBeEligible.tsx
@@ -16,6 +16,9 @@ export const WillBeEligible: React.VFC<{
   const apiTrans = getTranslations(tsln._language)
   const language = useRouter().locale as Language
 
+  const multipleOAS_GIS =
+    futureResults.filter((obj) => !!obj[Object.keys(obj)[0]]['oas']).length > 1
+
   return (
     <>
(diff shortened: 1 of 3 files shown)


components/ResultsPage/WillBeEligible.tsx
```diff
@@ -30,7 +33,29 @@ export const WillBeEligible: React.VFC<{
...
+
+        // show if some are non zero
+        const nonZeroExist = onlyOASGIS.some(
+          (key) => resultObj[age][key].entitlement?.result > 0
+        )
+
+        const enStr =
+          multipleOAS_GIS && nonZeroExist
+            ? partner
+              ? 'If your partner starts receiving at'
```
MarcoGoC, line 48: TODO  adding this ones to the messages

MarcoGoC (approved):
looks good

components/ResultsPage/WillBeEligible.tsx
alex-solo: Agreed! IF the decision is to keep things this way. But I think this is meant to be a temporary fix until we figure out a better way to display multiple OAS/GIS benefits or start accounting for that "years in Canada" thing that we talked about

PR #835, opened 2023-07-31
ADO 129729: Changed title of collapsed text and delete redundancies

alex-solo:
### Description

- Please see task, the requirements have changed since the task was first created. The message that "Individual incomes were used" is actually correct, but it was confusing since we are also display partner results in the same card for the future estimate of the client. So client "Will be eligible" amount is calculated using individual incomes but the current partner eligibility that is displayed in the card is calculated using combined incomes making the card "half right". Since the card has always been "user centric" that message is correct so a solution to this was to modify the language of the collapsed text title to "Some amounts were calculated based on individual income"

-
-


Maxim-Lam (approved):
Thanks for tackling this one Alex!

PR #836, opened 2023-08-01
hotfix for missing space

Maxim-Lam:
### Description

- Hotfix a space missing

-
- see desc

see task for params

### DIFF
--- a/components/ResultsPage/BenefitCards.tsx
+++ b/components/ResultsPage/BenefitCards.tsx
@@ -270,7 +270,7 @@ export const BenefitCards: React.VFC<{
       }
     } else if (benefitKey === BenefitKey.alws) {
       if (result.eligibility.result === ResultKey.ELIGIBLE) {
-        const ifYouApplyText = `${
+        const ifYouApplyText = ` ${
           apiTsln.detail.alwIfYouApply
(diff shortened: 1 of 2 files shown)


components/ResultsPage/BenefitCards.tsx
```diff
@@ -270,7 +270,7 @@ export const BenefitCards: React.VFC<{
       }
     } else if (benefitKey === BenefitKey.alws) {
       if (result.eligibility.result === ResultKey.ELIGIBLE) {
-        const ifYouApplyText = `${
+        const ifYouApplyText = ` ${
```
MarcoGoC: this `&nbsp;` isn't working ? 

```
alwIfYouApply:
      "If you apply, Service Canada will review your income tax return every year. You'll automatically be paid if your income is less than&nbsp;",
```

MarcoGoC (approved):
make sense but see comment

components/ResultsPage/BenefitCards.tsx
Maxim-Lam: good point. Space is actually missing from the end of the previous message "alwsApply". Reverting this change and adding it in the message

PR #837, opened 2023-08-02
fix message

Maxim-Lam:
### Description

- wrong message being displayed for ALW

- changed message being displayed
-

see ADO for params


MarcoGoC (approved):
it works as requested but I thing we should change the content of the card, the allowance is always for the partner.

PR #839, opened 2023-08-02
133872 - moving clawback to benefit

MarcoGoC:
### Description
- Move the message before the deferral options

- as per task


alex-solo (approved):
tested in dyno link and is looking like the message is moved to the correct place

PR #845, opened 2023-08-08
134696 - adding message and removing duplicate msg

MarcoGoC:
### Description
- 127979 changed partner to client but changing it back created duplicate messages
- the second message comes from the future card messages
- added a function to remove those duplicates 

- as above


Maxim-Lam (approved):
looks good

PR #858, opened 2023-08-16
 testing same name for secrets

MarcoGoC:
### Description
- updating the ingress to remove the branch name for the secrets, as proposed by Eric 

- the yarn change was necessary to chalk to work in pre-commit hook


P0NDER0SA (approved):
This should do the trick.  approved.

PR #859, opened 2023-08-17
moved sorting logic out of index page into utils

Maxim-Lam:
### Description

- remove sorting logic from front end 

- move the sorting logic from the front end to the back end
-


MarcoGoC (approved):
looks ok

PR #863, opened 2023-08-22
updated breadcrumbs according to ADO specs

Maxim-Lam:
### Description

- update breadcrumb according to spec

- see description
- result may vary slightly from ADO spec only because those are the links displayed on the prod environment vs our dev environment. It's coded so that it will display according to the environment 

### Additional Notes
[AB#131761](https://dev.azure.com/VP-BD/d5ca3cd4-19cb-4c8c-b3e6-a7068f4677e4/_workitems/edit/131761)

### DIFF
--- a/components/Layout/index.tsx
+++ b/components/Layout/index.tsx
@@ -53,7 +53,7 @@ export const Layout: React.VFC<{
   }
 
   const breadcrumbs =
-    process.env.APP_ENV !== 'production'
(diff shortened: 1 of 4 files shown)


MarcoGoC (dismissed):
adding the extra breadcrumbs after the initial declaration creates a refresh (also when changing languages) maybe within useEffect ?

MarcoGoC (approved):
not working, as it refreshes after load

PR #871, opened 2023-08-28
ADO-137892: Prevent negative months

alex-solo:
### Description

- We were getting a condition when negative months of deferral are allowed. When zero, we can bypass the oas deferral and get the results for the 'no deferral', residency only scenario.

-
-

### Additional Notes
 [AB#137892](https://dev.azure.com/VP-BD/DECD/_workitems/edit/137892)


wassila-aouchiche (approved):
tested the fixed and it worked fine : 
image

PR #872, opened 2023-08-29
Fix for duplicate results

Maxim-Lam:
### Description

- Moved the logic outside of the the class into the utils

- call function to removeDuplicates before passing the array to the component

- remove duplicates from future Results array if duplicate from result array

- see description

see bug 137225 for params specific bug.


MarcoGoC (approved):
Looks good

PR #877, opened 2023-08-29
'Fixed the Error validation behaviour for year bug'

wassila-aouchiche:
### Description

-Fixed the Error validation behaviour for year bug

- Modified calculateAge function under utils class
-

### Additional Notes
[AB#138387](https://dev.azure.com/VP-BD/DECD/_workitems/edit/138387)

### DIFF
--- a/utils/api/helpers/utils.ts
+++ b/utils/api/helpers/utils.ts
@@ -168,7 +168,7 @@ function addKeyValue(obj, key, val) {
  * This can and will return a decimal value, such as "65.5"!
  */
 export function calculateAge(birthMonth: number, birthYear: number): number {
-  if (!birthMonth || !birthYear) return 0
+  if (birthMonth == null || birthYear == null) return null
 
   const today = new Date()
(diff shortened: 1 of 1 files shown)


utils/api/helpers/utils.ts
```diff
@@ -168,7 +168,7 @@ function addKeyValue(obj, key, val) {
  * This can and will return a decimal value, such as "65.5"!
  */
 export function calculateAge(birthMonth: number, birthYear: number): number {
-  if (!birthMonth || !birthYear) return 0
+  if (birthMonth == null || birthYear == null) return null
```
MarcoGoC: why == instead of === ? 
I wasn't able to reproduce the original error

MarcoGoC (approved):

utils/api/helpers/utils.ts
wassila-aouchiche: let me change it to the strict one

PR #881, opened 2023-09-06
139376 new question

MarcoGoC:
### Description
- When user selects Yes to Do you already receive the Old Age Security pension? Change the 3rd residence question. Use number provided minus deferral time to calculate when years of residence were frozen. 

- as per task


alex-solo (approved):
let's try that!

PR #884, opened 2023-09-13
ADO-140726: Lock residence for scenarios when there is no current estimate, only …

alex-solo:
…multiple future estimates## [NNN](https://dev.azure.com/VP-BD/DECD/_workitems/edit/NNN) (ADO label)

### Description

- When there is no current estimate but estimates for some future years (for example 65 and 71), when calculating the OAS/GIS for 71, the residency should be locked at years of residency when they would start receiving at 65.


MarcoGoC (approved):
approving without testing

PR #887, opened 2023-09-13
ADO-140603: Only change residency if not only lived in Canada

alex-solo:
### Description

- We shouldn't change the years in Canada if the user has always lived in Canada when accounting for OAS deferral. 

For example, if they always lived in Canada, are 69 and deferred for 3 years, the residency used to calculate their OAS at 65 (when they became eligible) should be 40, not 37.

### DIFF
--- a/utils/api/benefitHandler.ts
+++ b/utils/api/benefitHandler.ts
@@ -398,7 +398,6 @@ export class BenefitHandler {
       )
     }
 
-    console.log('this.input.client', this.input.client)
     const clientOasNoDeferral = new OasBenefit(
       this.input.client,
       this.translations,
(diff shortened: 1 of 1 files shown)


utils/api/benefitHandler.ts
```diff
@@ -383,7 +383,7 @@ export class BenefitHandler {
     // Check OAS. Does both Eligibility and Entitlement, as there are no dependencies.
     // Calculate OAS with and without deferral so we can compare totals and present more beneficial result
 
-    if (this.input.client.receiveOAS) {
+    if (this.input.client.receiveOAS && !this.input.client.livedOnlyInCanada) {
```
MarcoGoC, line 386: wasn't there a case for which we removed the condition?

MarcoGoC (approved):

utils/api/benefitHandler.ts
alex-solo: Talked to Vero - we will accept this change and trust the automated tests to expose any potential bugs that we brought back by making this change.

PR #891, opened 2023-09-20
Updated to 2023 Q3 rates

alex-solo:
### Description

- Reran script with new values to create a new JSON file that the app will pull from.


MarcoGoC (approved):
verified, thanks

PR #896, opened 2023-09-22
140928 65 estimate not provided

numbap:
### Description

- Users should not be able to claim OAS until the month after they turn 65

- Remove the ability to claim OAS if age is beloe 65 + 1 month. 
-

Enter an age of exactly 65 and 0 months. The form should not ask you if you currently collect OAS. 

### Additional Notes

Another bug to be addressed in the future is that deferrals are off by one month due to this glitch.

### DIFF
--- a/__tests__/pages/api/expectUtils.ts
+++ b/__tests__/pages/api/expectUtils.ts
@@ -182,7 +182,7 @@ export const income10k = {
 }
 
 export const age65NoDefer = {
-  age: 65,
+  age: 65.08,
   oasDefer: false,
   oasAge: undefined,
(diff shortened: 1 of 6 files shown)


wassila-aouchiche (commented):
if we changed the age from 65 to 65.08 in the benefit handler class how about the other classes that we are using the 65 age to validate some cases ? for example FutureHandler class

MarcoGoC:
We probably need more info about this one, from reading the task, the changes are not what it was requested

MarcoGoC (changes requested):
cannot reproduced

PR #913, opened 2023-10-03
Future Ages based on residency (allow partner residency < 10)

alex-solo:
### Description

- This addresses scenarios where partner residency is under 10
- **DONE** 2 tests to take a look at CALC-145 and CALC-156 before we can merge.

- adjust tests
- also includes bug fix, partner should not be eligible for ALW at 63 with 9 years residency.

### Additional Notes

More info in ADO-141941


MarcoGoC (commented):
image
dates Oct 1951, and Oct 1960. 
according to the test is missing a paragraph for client at age 74.

alex-solo:
@MarcoGoC Added some fixes. I think I Cmd+Z too many times and initially erased the code that made the partner not eligible for ALW, so that fix is back! And what @VeroBr0 said about the test, it was indeed a false positive. I didn't see that we have a reason "Not enough years", which I now added as an expected value in CALC-156. Let me know if it's working as it should, amounts at 74 for client aren't showing up since they're the same as Vero mentioned.

VeroBr0:
Did you want to do ALW not enough years or OAS not enough years? This did make me notice we take the reason as a parameter when it's gisNotEligible but we don't for OAS. I can update that in an upcoming cleanup, but for now calling it directly like you did works.

MarcoGoC:
I see the changes but not sure how to test it.

MarcoGoC (approved):
looks good

__tests__/pages/api/oasDeferral.test.ts
```diff
@@ -699,7 +699,7 @@ describe('OasDEferral', () => {
     //partner results
     expectOasNotEligible(res, true)
     expectGisNotEligible(res, ResultReason.OAS, true)
-    expect(res.body.partnerResults.oas.eligibility.reason).toEqual(
+    expect(res.body.partnerResults.alw.eligibility.reason).toEqual(
```
MarcoGoC, line 702: missed that

PR #916, opened 2023-10-04
143253-Fix empty deferral heading showing up

Maxim-Lam:
### Description

- fix empty deferral heading showing up when no deferral is being shown

-
-

- test for age < 65 & income > max threshold (no heading, original bug)
- test regular deferral

### DIFF
--- a/utils/api/benefits/oasBenefit.ts
+++ b/utils/api/benefits/oasBenefit.ts
@@ -578,7 +578,9 @@ export class OasBenefit extends BaseBenefit<EntitlementResultOas> {
     ) {
       // your Deferral Options
 
-      text += `<p class='mb-2 mt-6 font-bold text-[24px]'>${this.translations.detail.yourDeferralOptions}</p>`
+      if (this.deferral) {
+        text += `<p class='mb-2 mt-6 font-bold text-[24px]'>${this.translations.detail.yourDeferralOptions}</p>`
+      }
 
       // if income too high
       if (this.eligibility.reason === ResultReason.INCOME) {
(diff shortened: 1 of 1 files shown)


MarcoGoC:
this is from develop 
image
your branch is missing the header

MarcoGoC (changes requested):
see comment

Maxim-Lam:
> this is from develop image your branch is missing the header

The parameters for you example do not match the conditions of the bug. The bug occurred when user is < 65 and income > max threshold. Let me check to see what should be displayed in under these conditions. @MarcoGoC

MarcoGoC:
I know, but your change breaks all other scenarios

utils/api/benefits/oasBenefit.ts
```diff
@@ -578,7 +578,9 @@ export class OasBenefit extends BaseBenefit<EntitlementResultOas> {
     ) {
       // your Deferral Options
 
-      text += `<p class='mb-2 mt-6 font-bold text-[24px]'>${this.translations.detail.yourDeferralOptions}</p>`
+      if (this.deferral) {
```
MarcoGoC: this breaks all the other scenarios

Maxim-Lam: @MarcoGoC I moved the heading into the respective if statements. I am unable to see of a way that we can append the heading outside of the two if statements without breaking other scenarios. Open to suggestions.

MarcoGoC (approved):
not ideal but this will work for all cases

PR #919, opened 2023-10-04
143862 - Refactor - moving InvSeparated out

MarcoGoC:
### Description
- Refactor, moving Inv Separated cases out of the benefit Handler 

- As above

- Passes all the current tests


alex-solo (approved):
Thank you Marco, this reduces the benefitHandler by a LOT!

PR #922, opened 2023-10-05
142179 - Refactor Alert and Heading

shawn320:
### Description

- Remove DS Heading
- Add ContextualAlert component to replace DS

The ContextualAlert is imported on index.tsx as Message. Try to change `type` prop to check different type of alert. There are 4 types: info, success, warning, danger. Different type should show different icon and color.

EE was using an outdated version of the DS Alert, to checkout latest design, visit [Figma](https://www.figma.com/file/Q9p7J0SCI3Xn5M1QUVdiT2/OAS-Benefits-Estimator?type=design&node-id=9208-138749&mode=design&t=ZzL0PIj2DeOPP6uA-0)


components/Layout/index.tsx
```diff
@@ -132,20 +129,20 @@ export const Layout: React.VFC<{
               logoAltText: tsln.logoAltText,
             }}
           />
-          <Heading
+          <h1
             id="applicationTitle"
-            title={title}
-            className="mb-8 mt-4 sm:mt-12 sm:w-[100%]"
-          />
+            className="font-header-gc font-[700] text-[#333333] text-[38px] leading-[42px] mb-8 mt-4 sm:mt-12 sm:w-[100%]"
```
MarcoGoC: globals.css has a class  .h1  aren't we duplicating code?  is a question

MarcoGoC: Questions page and ErrorsSummary also use the message component, maybe added them to this PR

components/Layout/index.tsx
```diff
@@ -129,10 +129,7 @@ export const Layout: React.VFC<{
               logoAltText: tsln.logoAltText,
             }}
           />
-          <h1
-            id="applicationTitle"
-            className="font-header-gc font-[700] text-[#333333] text-[38px] leading-[42px] mb-8 mt-4 sm:mt-12 sm:w-[100%]"
-          >
+          <h1 id="applicationTitle" className="h1 my-8">
```
MarcoGoC, line 132: Looks good to me, just check with Andreanne about the font size
this is from prod

image

MarcoGoC (approved):
see comment

components/Layout/index.tsx
MarcoGoC: versus
image

shawn320: > Andreanne

H1 font size in mobile is different, updated.

PR #931, opened 2023-10-12
'Fixed the issue with the Age error message'

wassila-aouchiche:
### Description

-  Added a custom validation to joi schema validation for the age

- Updated schema.ts file

- Tested locally with different scenarios (age under 18, birth year between 1800 and 2005 and greater than the current year which is 2023)


MarcoGoC (approved):
works

PR #932, opened 2023-10-13
ADO-140665: fix involuntary separated scenarios

alex-solo:
### Description

- The key to this solution was fixing how we determine whether we use rate table 1 (single income) vs rate table 2 or 3 which correspond to combined incomes.

```const useT1versusT2_T3 =
        applicantGisResultT1 + partnerGisResultT1 >
        applicantGisT2_T3 + partnerGisResultT2
```

- Another set of scenarios failed as a result, a lot of which were false positives before so those tests have been fixed 
- Uncommented a few tests which also failed but some of those were due to incorrect ages.

### DIFF
--- a/__tests__/pages/api/gisCoupleAlwBenefit.test.ts
+++ b/__tests__/pages/api/gisCoupleAlwBenefit.test.ts
@@ -4,946 +4,976 @@ import {
   ResultReason,
 } from '../../../utils/api/definitions/enums'
 import { getTransformedPayloadByName } from '../../utils/excelReaderUtil'
-import { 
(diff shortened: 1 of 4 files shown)


MarcoGoC (approved):
changes makes sense but I do not get the same values as the task. so i'll be this to Vero and the CALC test

alex-solo:
> changes makes sense but I do not get the same values as the task. so i'll be this to Vero and the CALC test

Thanks Marco. Some of the values in the tests were found to be incorrect -  Lorelei has made corrections on the excel sheet in red.

PR #936, opened 2023-10-17
146402 - Adding living country to ALW benefit

MarcoGoC:
### Description
- ALW was being calculated for when partner is not eligible for GIS since is not living in Canada

- Added LivingCountry to AlwBenefit Class


alex-solo (approved):
Looks good, thanks Marco

PR #939, opened 2023-10-23
145339-fix null error when selecting when country not canada

Maxim-Lam:
### Description

- added variable for yearsInCanadaSince18 to OASBenefit since it is lost when newInput is passed

- see description
-

See bug 145339 for steps


MarcoGoC:
As per slack comments I am not sure this solution is correct

PR #943, opened 2023-11-03
145339-fix-null-deferral

Maxim-Lam:
### Description

- fixed null values in oas deferral text and table not appearing
- tests no longer failing

- properly pass in living country and lived only in canada values to the OasEligibility
-

see ado for parameters


alex-solo (approved):
perfect!

PR #945, opened 2023-11-06
July 2013 OAS deferral

alex-solo:
### Description

Long awaited July 2013 OAS deferral logic. See associated ADO task to get a full idea of requirements. In brief, there are front and "back" end portions to this.

**Front-end**: When client selects "Yes", they already receive OAS, the deferral dropdown values are populated based on their age in July 2013. If they were between the ages 65-70 at that time, they could have deferred by 70 - age in July 2013. 

**Back-end**: If the client answers "No" to whether they already receive OAS, we need to know what the users age was in July 2013 and based on that calculate the "deferred" scenario (as a reminder, we calculate the deferred and non-deferred scenarios and compare which would provide a more beneficial estimate for the user). The comparison logic is still the same, we just needed to calculate the deferral scenario differently - keeping in mind the "true" number of months the client could have deferred based on the fact that OAS deferral only came into effect in July 2013.

- merge Vero's branch
- front-end July 2013 OAS deferral logic
- back-end July 2013 OAS deferral logic
- new test scenarios
- test utils logic to calculate age in July 2013 based on birth year and month
- upload latest scenarios excel sheet from sharepoint

back-end is hard to test, the automated tests do that but should test the front end dropdowns making sure the correct number of years/months are being populated. For example, if the user was over 70 in July 2013, shouldn't have the option to select anything, if they were 67.5, should be able to select up to 2 years and 6 months


Maxim-Lam (approved):
Everything looks good, Alex. I tested with the scenarios you mention and everything works like you describe.

PR #950, opened 2023-11-17
ADO-153380: Legal status "No" returned empty age array

alex-solo:
- Need to ensure that when partner legal status is "No", we don't pass a NaN to function that determines "future ages". 
- Client legal status "No" results in user not being able to continue so we don't need to worry about that case.


MarcoGoC (approved):
+1 thanks

PR #954, opened 2023-11-28
153393 - changed date to 1900 to match error message

MarcoGoC:
### Description
- Changed the date validation from 1800 to 1900 to match the error message

- as above 

- enter a year before 1900


Maxim-Lam (approved):
Looks good

PR #956, opened 2023-11-30
157393 - Check Benefit exists

MarcoGoC:
### Description
- when checking for oas which is not part of the array it crashes, by adding a condition we avoid that 

- Adding a condition to check for non undefined values

- Go to staging or Prod try the values. 
- Try the same on this branch 

### Additional Notes
-


Maxim-Lam (approved):
Nice fix

PR #957, opened 2023-11-30
156840 - Bug fix on What you told us section

shawn320:
### Description

- First letter of category name should be in capital. 

- Go through the quesions to go to the result page, check "What you told us" section, In moile view, the first letter for each category should be in capital.


MarcoGoC (approved):
Looks good, however it could be refactor to use the `utils/api/definitions/fields  fieldDefinitions ` where the key and category are already defined and the hardcoded categoryMapping can be removed

MarcoGoC (approved):
Thanks

PR #960, opened 2023-12-05
ADO 134595: Refactor benefitHandler

alex-solo:
### Description

- Refactored BenefitHandler class to take field related concerns out and keep benefit logic only.

- create FieldsHandler class

- Run through the app, try some examples, make sure the fields are populated, appear/disappear based on previous inputs and make sure the results page comes up and displays results as it should

### Additional Notes

- We could refactor further into "InputsHandler" and "TranslationHandler" but this refactor gives us a much more concise file and might be good enough of a refactor. Can discuss with team later, refactor as needed during other tasks.


MarcoGoC (approved):
Looks good to me, and passes all current tests

PR #966, opened 2023-12-11
156831 - updated link anchor for footer

MarcoGoC:
### Description
- navigate using Tab to the top of the page where there is a hidden link for "Skip to about government" on click it will move the page to the footer 

- as above


Maxim-Lam (approved):
Hey Marco, everything looks good on Chrome but have you tested on FF? It does not seem to tab properly for me

PR #976, opened 2023-12-15
Fix 143948

Maxim-Lam:
### Description

Fix for 143948. Only display the message if the age used to calculate the benefit is not the age that was input and if lived in canada is less than 40. We don not want to display 40 years in the message

- pass the query age in from future handler down to oas benefit
- add conditions to display appropriate message (less than 40 in canada, age != input age)
- break message down to display appropriately

- check bug for parameters

### Additional Notes

-

### DIFF
--- a/utils/api/benefitHandler.ts
+++ b/utils/api/benefitHandler.ts
@@ -33,19 +33,19 @@ export class BenefitHandler {
   fields: FieldsHandler
   future: Boolean
   compare: Boolean
-  age: number
+  formAge: number
 
   constructor(
     readonly rawInput: Partial<RequestInput>,
     future?: Boolean,
-    age?: number,
+    formAge?: number,
     compare: Boolean = true
   ) {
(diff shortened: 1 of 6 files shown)


utils/api/benefitHandler.ts
```diff
@@ -158,7 +161,7 @@ export class BenefitHandler {
       false,
       this.future,
       false,
-      this.input.client.age
```
alex-solo, line 161: in cases when age from the form is different from age we want to calculate the benefits for (ex. form input 61, we run the benefit handler for both 61, then future ages, for example 65, 72 etc.) we would need to use right age. With the change you added here, it would be 61 for every case.

alex-solo (changes requested):

utils/api/benefitHandler.ts
Maxim-Lam: fixed

alex-solo (approved):

PR #978, opened 2023-12-15
Pipeline updates

andr0272:
I added the workspace `oas/` into the tag instead of the repository name in my last commit. This cleans that up and shortens the commit hash to the first 7 digits.


MarcoGoC (approved):
:crossed_fingers:

PR #991, opened 2023-12-29
144220 Income questions updates

shawn320:
### Description

Added a new field for salary exemption under Income category. 

Calculation updates:
To calculate income for GIS, ALW and ALWS, take the input from the first field and SUBTRACT these amounts from the second field:
the first $5,000
50% of next $10,000
The maximum subtraction possible is $10,000 if input is $15,000 or more.

To calculate income for partner GIS, ALW and ALWS, apply the same calculation to the inputs.

To compare against recovery threshold for OAS pension, use income from first field only (no subtractions).

Updated error handling: Second income field cannot be greater than net income, also apply to partner income.

Updated income question: The income question, hint expand text, and error message should change dynamically based on user's input on the receiveOAS field. More detail on [Figma](https://www.figma.com/file/Q9p7J0SCI3Xn5M1QUVdiT2/OAS-Benefits-Estimator?type=design&node-id=24744-68515&mode=design&t=t0mF4zavTdmBHWZu-0)


utils/api/helpers/fieldClasses.ts
```diff
@@ -31,6 +33,24 @@ export class IncomeHelper extends FieldHelper {
         (maritalStatus.partnered && partner !== undefined))
   }
 
+  /**
+   * Calculate the income after applying deductions.
+   * Deduct the first $5,000 entirely, then 50% of the next $10,000.
+   */
```
MarcoGoC, line 39: is this twice?

components/ResultsPage/YourAnswers.tsx
```diff
@@ -10,6 +10,7 @@ import { MonthsYears } from '../../utils/api/definitions/types'
 import { Accordion } from '../Forms/Accordion'
 import { fieldDefinitions } from '../../utils/api/definitions/fields'
 import { FieldCategory } from '../../utils/api/definitions/enums'
+import { number } from 'joi'
```
MarcoGoC: not sure about this, needs to be doublecheck, the answers are all string

components/ResultsPage/YourAnswers.tsx
```diff
@@ -41,6 +42,20 @@ export const YourAnswers: React.VFC<{
...
+  function salaryExemption(income: number): number {
+    if (income <= 5000) {
+      return income
+    } else if (income <= 15000) {
+      return 5000 + (income - 5000) * 0.5
+    } else {
+      return 10000
+    }
+  }
+
```
alex-solo, line 56: "Your answers" should just capture inputs from the questions page, why the income logic here?

alex-solo (changes requested):
The calculations themselves seem to be working according to the new logic, nice! The tests are failing because both new fields are required as per Joi schema but the tests suite does not provide those values in any of the scenarios.

You can probably address this by setting it to zero if 'undefined' when the test suite excel sheet is being ingested. OR, since we'll need different new scenarios in the test suite, we'll need a new column for this new input. All the current tests can be auto-filled to zero or "N/A" as long as the test scripts know how to read the sheet, it's all fair game.

Take a look at excelReaderUtil.ts to see how the sheet is being processed currently. Let me know if you have any more questions!

components/ResultsPage/YourAnswers.tsx
shawn320: The "What you've told us" section has a new field called Salary exemption, this field does not display the original input value for `incomeWork` field, instead it displays the deducted value after calculation, so I added this to adjust the `incomeWork` value.

components/ResultsPage/YourAnswers.tsx
shawn320: imported by accident, removed.

utils/api/helpers/fieldClasses.ts
shawn320: replied in Alex's comment

alex-solo (approved):
already approved based on previous conversations

PR #993, opened 2024-01-08
Jump to error field when submitting faulty form

Maxim-Lam:
### Description

- When trying to submit a faulty form, bring user to the error field

- added id to error field
- jump to the error field if form is not valid

- fill out form, change a field to have an error, click estimate my benefit.
- it should bring you back to the error field

### Additional Notes

-


alex-solo (approved):
nicely done. If the requirement was to bring the focus to the summary of errors and not the first occurrence of error. Also, as a user, would prefer a gentle scroll up instead of snapping into focus but if it's all good with Andreanne, then great!

PR #995, opened 2024-01-15
ADO-164296: Add attributes as they are and expand them only in the button component

alex-solo:
### Description

- We weren't getting the attributes down to the button component


Maxim-Lam (approved):
LGTM

PR #1000, opened 2024-01-18
Remove null values before returning array

Maxim-Lam:
### Description

- application error given certain combination of parameters with widowed marital status. Client result array would return a null value and app would crash

- remove null values from array before returning it 

- see bug 165834

### Additional Notes

-


alex-solo (approved):
Gets rid of the null, but consider chaining filter after map so everything is neatly tied to the same variable

PR #1002, opened 2024-01-23
149553-gap-in-residency

Maxim-Lam:
### Description

-  WIP PR still testing

- if living country is not CAN, use actual years in canada in calculations

-

### Additional Notes

-


alex-solo (approved):
Approving so these can be tested in the dev link

alex-solo (approved):
Approving so these can be tested in the dev link

PR #1008, opened 2024-01-31
Fix missing next step

Maxim-Lam:
### Description

- fix missing next step

- added missing next step title to alw next steps

- see bug for params

### Additional Notes

-


alex-solo (changes requested):
to delete console.log

alex-solo (approved):

PR #1012, opened 2024-02-06
ADO 168576 - Appropriate markup to fix accessibility issues on questions and results pages

alex-solo:
### Description

- Marital status options should be a list (was using p tags)
- bold text should be in a strong element for readers to know it's impportant

### DIFF
--- a/i18n/web/en.ts
+++ b/i18n/web/en.ts
@@ -308,7 +308,7 @@ const en: WebTranslations = {
   resultsPage: {
     header: 'Table of estimated monthly amounts',
     general:
-      'The following is only an estimate of your eligibility and monthly payments <span style="font-weight: bold;">based on current rates</span>. Amounts may increase with the cost of living. Changes in your circumstances may also impact your results.',
+      'The following is only an estimate of your eligibility and monthly payments <strong style="font-weight: bold;">based on current rates</strong>. Amounts may increase with the cost of living. Changes in your circumstances may also impact your results.',
     onThisPage: 'On this page',
     tableHeader1: 'Benefit',
     tableHeader2: 'Estimated monthly amount (CAD)',
--- a/i18n/web/en.ts
+++ b/i18n/web/en.ts
@@ -308,7 +308,7 @@ const en: WebTranslations = {
   resultsPage: {
     header: 'Table of estimated monthly amounts',
     general:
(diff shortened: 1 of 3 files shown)


i18n/web/en.ts
```diff
@@ -308,7 +308,7 @@ const en: WebTranslations = {
   resultsPage: {
     header: 'Table of estimated monthly amounts',
     general:
-      'The following is only an estimate of your eligibility and monthly payments <span style="font-weight: bold;">based on current rates</span>. Amounts may increase with the cost of living. Changes in your circumstances may also impact your results.',
+      'The following is only an estimate of your eligibility and monthly payments <strong style="font-weight: bold;">based on current rates</strong>. Amounts may increase with the cost of living. Changes in your circumstances may also impact your results.',
```
Maxim-Lam: Doesn't <strong> tag already make font-weight: bold? might be a redundancy?

i18n/web/fr.ts
```diff
@@ -312,7 +312,7 @@ const fr: WebTranslations = {
   },
   resultsPage: {
     header: "Tableau des résultats d'estimation",
-    general: `Les résultats suivants ne sont qu'une estimation de votre admissibilité et de vos paiements mensuels <span style="font-weight: bold;">basée sur les montants actuels</span>. Ceux-ci peuvent augmenter avec le coût de la vie. Des changements dans votre situation pourraient aussi modifier vos résultats`,
+    general: `Les résultats suivants ne sont qu'une estimation de votre admissibilité et de vos paiements mensuels <strong style="font-weight: bold;">basée sur les montants actuels</strong>. Ceux-ci peuvent augmenter avec le coût de la vie. Des changements dans votre situation pourraient aussi modifier vos résultats`,
```
Maxim-Lam: Same comment as above

Maxim-Lam (changes requested):
Looks good. Just the one comment to consider.

i18n/web/en.ts
Maxim-Lam: sorry, wrote "<" strong ">" tag*

i18n/web/fr.ts
alex-solo: good catch, got rid of styles

Maxim-Lam (approved):
Looks good!

PR #1022, opened 2024-02-19
Use max oas instead of oas+gis

alex-solo:
### Description

- Using different logic to calculate the best scenario - only use OAS. Whichever is higher (deferred or non-deferred)

- logic change
- test scenarios expected results change

- n/a. All tests passing

### Additional Notes

-


MarcoGoC (approved):
without testing it

PR #1025, opened 2024-02-20
177286 - Remove max income validation

MarcoGoC:
### Description
- Remove error message 2 income (work) cannot be bigger than income 

- as above

- Enter a 'work or self-employee' value bigger than income, an error message should not be shown 

### Additional Notes
-


Maxim-Lam (approved):
Looks good

PR #1028, opened 2024-02-27
168542 - Radio button fix, maybe

MarcoGoC:
### Description
- Error messages showed when selecting No to Citizenship question is not read by an screen reader 

- Hopefully the change is read by the screen-reader

- Need a screen-reader

### Additional Notes
-


alex-solo (approved):
Approving without having tested with screen reader. I believe Andreanne will be able to test it

PR #1031, opened 2024-03-15
175541 - Remove Countries

MarcoGoC:
### Description
- removed and adjusted tests for countries errors

- added Outside of Canada to list of countries used for the validation

- on staging enter chose Outside of Canada to get the error. then run this branch to not get the error

### Additional Notes
-


alex-solo (approved):
I thought the coutries were already removed in a previous PR recently? Looks like they're back but being commented out. Is there a reason for that?

PR #1037, opened 2024-03-27
bug  189595  fix

MarcoGoC:
### Description
- For over 70yrs old on 2013,  Deferral is not asked and the validation crashed expecting a value. this fixes the issue  

- parsing only when the value is available, when is not default to zeros 

- see task

### Additional Notes

-


Maxim-Lam (approved):
Looks good. Didn't test

PR #1043, opened 2024-04-10
ADO-171109: Fixed SEO issue and check if alpha/beta site to add correct breadcrumbs

alex-solo:
### Description

- In addition to the ADO task, I think we were missing out on some SEO for the alpha site. Meta data links in the head of the HTML were pointing to the wrong place.

- Breadcrumbs for alpha site should also be the same ones we see in dev. Only beta site has the "complete" breadcrumbs.

-

### Additional Notes

-


MarcoGoC (approved):
+1

PR #1046, opened 2024-04-18
120732 - Inv separated changes

MarcoGoC:
### Description
- new text for question and question help in both languages 

- as above per task

- match new text against the task

### Additional Notes
-


Maxim-Lam (approved):
looks good

PR #1047, opened 2024-04-18
181300 - overview changes

MarcoGoC:
### Description
- Updated text as per task and Figma

- as above

- verified text against figma


alex-solo (approved):
Approving for QA to test to make sure everything is in the right place but looks good from here.

PR #1050, opened 2024-04-26
181300 - overview css changes

MarcoGoC:
### Description
- As per description on the task comments, missing french punctuation, and css changes to look alike the canada.ca suggestion. 

- as above

-

### Additional Notes

-


alex-solo (approved):
Only thing I can think would be better is to move colours into a tailwind config file for better reuse, otherwise, looks like it's doing what is required.

alex-solo (approved):
Only thing I can think would be better is to move colours into a tailwind config file for better reuse, otherwise, looks like it's doing what is required.

PR #1052, opened 2024-05-13
185038 income labels

MarcoGoC:
### Description
- As per task

- see task

- run and compare changes to the task requests

### Additional Notes
-


alex-solo (dismissed):
image

The work income question is optional, but it should prefill or automatically fill with 0 and allow the user to click "Next" to proceed to the next card.

alex-solo (approved):
image
 
It's still a bit confusing for the user if they delete the input and it's empty, they cannot proceed. But maybe a question for Andreanne about the specific behaviour that is required.

PR #1056, opened 2024-05-16
186058 - adding tooltip and question help for Residence

MarcoGoC:
### Description
- Adding Residence Tooltip and QuestionHelp 

- see above 

- Compare test from task to localhost text

### Additional Notes
-

### DIFF
--- a/i18n/api/en.ts
+++ b/i18n/api/en.ts
@@ -149,13 +149,13 @@ const en: Translations = {
     [FieldKey.INCOME_AVAILABLE]:
       'Providing your income will give you more accurate results.',
     [FieldKey.INCOME]:
-      'Add all types of income after deductions, including: pensions, benefits, annuities, salaries and retirement fund withdrawals (you can refer to line&nbsp;23600 of your T1). Do not include what you receive from any of the Old Age Security benefits.',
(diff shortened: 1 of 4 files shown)


i18n/api/en.ts
```diff
@@ -169,6 +169,10 @@ const en: Translations = {
       'Only count the number of years since the age of 18.',
     [FieldKey.PARTNER_YEARS_IN_CANADA_SINCE_18]:
       "If you're not sure of the exact number, you may enter an estimate.",
+    [FieldKey.LIVED_ONLY_IN_CANADA]:
+      'Do not include periods when you were outside Canada for at least 6 months at a time. Some exceptions apply, such as working for a Canadian employer abroad.',
+    [FieldKey.LIVED_ONLY_IN_CANADA]:
```
alex-solo: These are duplicated field keys. I'm guessing the other one is supposed to be for partner

alex-solo (dismissed):
Noticed duplication in en.ts and fr.ts files

i18n/api/en.ts
MarcoGoC: good catch, that's correct

alex-solo (approved):
looks good!

PR #1067, opened 2024-06-05
dont round the future age to get accurate age at first eligibility

alex-solo:
### Description

- This ADO task was already completed under a different name ("residency under 10 years") but this PR is simply a bug fix that leaves the ages not floored so we can know precisely the age when someone becomes eligible with years and months.


MarcoGoC (approved):
not exactly sure what is this lol

PR #1073, opened 2024-06-07
bug fix for ADO-133070

alex-solo:
### Description

- When a client is not yet eligible because they are under eligibility AGE (but have sufficient residency years at time of eligibility), months should not be taken into account. For example, if a client is 63.8 years old and have lived in Canada all their life, their age of eligibility should be 65. However, if for the same age, their residency is only 7 years in Canada, by the time they are eligible for OAS (at 10 years of residency), we should take months into account and so the client will be 66.8 years old at time of 10 years of residency.

### DIFF
--- a/utils/api/futureHandler.ts
+++ b/utils/api/futureHandler.ts
@@ -53,8 +53,6 @@ export class FutureHandler {
       String(this.query.livingCountry)
     )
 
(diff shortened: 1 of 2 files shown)


MarcoGoC (approved):
part of the one from yesterday? i thought that there was something missing  :)   :+1:

PR #1077, opened 2024-06-10
Dynamic links YAML: identity assign if does not exist

alex-solo:
### Description

- Trying a new approach: run identity assign if the assignment step hasn't occurred yet, and don't if it already does.

-

-

### Additional Notes

-


MarcoGoC (approved):
:crossed_fingers:

PR #1085, opened 2024-06-18
219714 - Updated css for visited links

MarcoGoC:
### Description
- The main purpose was visited links, however as part of those changes many, options had to be modified
- All links have a class that adds the svg for OpenNewTab 
- adding missing aria-label for 'open an new tab' 

- as described above

- dl and run 

### Additional Notes
-


alex-solo (approved):
lots of cascading changes, let's test once merged.

PR #1115, opened 2024-11-19
Browser navigation and saving results for results page

alex-solo:
### Description

- Back button navigation using query strings
- Modify form class to write to session storage (response and the inputs) for results page to use. Before, the results page re-did the calculation.


Maxim-Lam (approved):
Approved

PR #1124, opened 2024-11-28
Fix results page on language toggle

alex-solo:
### Description

- Because some of the text and card details is stored in the calculated results object, we have to run the calc again when the language is being toggled to "refresh" it to the correct language


MarcoGoC (approved):
good job, thanks

PR #1141, opened 2025-02-04
Fix key error

MarcoGoC:
### Description
- As above 

- As above

- run locally the 'new-results' open the console make sure warnings are included see the error 
- runs this branch see error gone

### Additional Notes
-


alex-solo (approved):
thanks for fixing that

PR #1144, opened 2025-02-06
311221 focus modal

MarcoGoC:
### Description
- This could fix the issue of the focus on the page not on the modal, and maybe the read the modal, unknown until it is merge and check with nvda   

- as above 

- 

### Additional Notes
-


Maxim-Lam (approved):
+1

PR #1151, opened 2025-02-15
making null benefit object when partner entit = null

MarcoGoC:
### Description
- as stated above

- as above


alex-solo (approved):
let's try this and put it in Dev and test thoroughly

PR #1159, opened 2025-02-19
Spike related issues

alex-solo:
### Description

- See the ADO related to spikes: We want to not show certain results when the other (partner or user) is not eligible

- We can treat two cases separately, one where user is older than client and vice versa. Looking at the case of user older than client, we want to determine when their eligibility age is. While iterating over future age sets, we can check if it's the eligibility age and if the partner's results yield any results. If it's not eligibility age and no results from partner, then we don't need to show the results for the user either.

-

### Additional Notes

-

### DIFF
--- a/utils/api/futureHandler.ts
+++ b/utils/api/futureHandler.ts
@@ -190,24 +190,33 @@ export class FutureHandler {
   }
 
   private getPartneredResults() {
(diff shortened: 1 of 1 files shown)


MarcoGoC (approved):
lets try

PR #1166, opened 2025-02-21
Use original amounts in deferral table

alex-solo:
### Description

- ADO-321256
- bug QC found to cap deferral months when doing PSD calc

- We also need to cap years in Canada to 40 in all cases so changes to OasEligibility were made as well.

-

### Additional Notes

-


Maxim-Lam (approved):
good job

PR #1168, opened 2025-02-26
Keep the partner benefit and recalculate in special case

alex-solo:
### Description

- ADO-319323

- We need to recalculate parter benefits in a case when they are older than client and already eligible for OAS/GIS from previous age in order to force the calculation to use a different table to calculate the benefit. In case where the couple is involuntary separated we don't need to do this because the same table is used throughout.

-

### Additional Notes

-


MarcoGoC (approved):
to be tested

PR #1170, opened 2025-02-27
261959 - Income question update

MarcoGoC:
### Description
- As per task, income question is being overwritten according to the receive oas 

- as Above

- 

### Additional Notes
-


alex-solo (approved):
not tested but nothing jumping out at me.

PR #1173, opened 2025-03-03
ADO-315436: Show ALW for partner and fix future ages function

alex-solo:
### Description

- Simplified the agesInputObj and fixed function that was responsible for mapping and changing a variable - shouldn't use map here, changed to a reduce.

-

-

### Additional Notes

-


MarcoGoC (approved):
looks good

PR #1189, opened 2025-03-07
ADO-325430: Hide duplicate amount results

alex-solo:
### Description

- When we recalculate the amount for partner, if it's the same amount as they were previously receiving we will now hide it to avoid confusion in cases when this happens in the same year.

-

-

### Additional Notes

-


MarcoGoC (approved):
approved without testing

PR #1190, opened 2025-03-07
323839 - change select component

MarcoGoC:
### Description
- changing the select component to use the same for the app

- changing component


alex-solo (approved):
Should test in develop

PR #1194, opened 2025-03-12
323839 - adding value to restore previoulsy selected value

MarcoGoC:
### Description
- as per title 

- 

- go back and for between questions and results


components/Forms/Select.tsx
```diff
@@ -40,6 +40,8 @@ export const FormSelect: React.VFC<SelectProps> = ({
     (item) => item.key === field.value
   )
 
+  console.log('current item', currentItemKeyText)
```
alex-solo: to delete

alex-solo (approved):

PR #1207, opened 2025-03-19
323853 - Button using title instead

MarcoGoC:
### Description
- As per request using title instead of alt.  

- added title to the Button Props


Maxim-Lam (approved):
Good

PR #1220, opened 2025-03-24
Refactor remove element

alex-solo:
### Description

- Changing the previous change so that we can avoid using useEffect entirely and go with the React way using actual state

-

-

### Additional Notes

-


Maxim-Lam (approved):
Looks good!

PR #1240, opened 2025-06-12
Fix PSD resetting OAS deferral amount

Maxim-Lam:
### Description

- fix for bug, tested against excel smk test 16-19

-

-

### Additional Notes

-


alex-solo (approved):
Test against cases where criteria is met but something else is different like marital status or living outside of Canada for example

